PR #9, opened 2015-06-22
Add `freeUserPan` prop

Chrisui:
Currently when the map component enters the "receive props" lifecycle it will nuke any user pan position as it determines the provided `props.center` to be different from the current center in the geo service.

Sometimes you won't want this behaviour as you want to allow users to pan freely while the component is updated.

As a simple solution I've added a new `freeUserPan` prop. We should also consider exposing an `onPan` hook so we can keep state higher and simply pass the new center down. For now I think the new prop solution is simplest.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -62,7 +62,8 @@ export default class GoogleMap extends Component {
     hoverDistance: PropTypes.number,
     debounced: PropTypes.bool,
     margin: PropTypes.array,
-    googleMapLoader: PropTypes.any
+    googleMapLoader: PropTypes.any,
+    freeUserPan: PropTypes.bool
   };
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


Chrisui:
Just noticed I need to check for nextProps.center as well so definitely wait for that :)

istarkov:
Dear @Chrisui can you explain me (or show an example)  in what situation map center is not the same as user map position? (*for example if now i move the map and some event cause redraw - all is fine*)

Mb via skype (where we can share screen) to save your time.  My is `cybice`.
(It looks like i have some translations or misunderstanding problems)

Chrisui:
Hey @istarkov sorry for not being too clear! 

Here is a demo of the problem. http://chrisui.github.io/twitter-viz/ (ignore the request for location it doesn't matter)

If you pan the map and then click the button in the top left you see it pans back to it's original `center` position.

You can see the code here: https://github.com/Chrisui/twitter-viz/blob/gh-pages/client.js#L67

Note how the button click does nothing but change state (causing update) which passes a new dummy prop to the map to put it into `componentWillReceiveProps` lifecycle. The [logic handling this lifecycle](https://github.com/Chrisui/google-map-react/blob/master/src/google_map.js#L416) within this library sees that the props center does not match the user panned center and so pans back to the props center.

This is the issue this pull request solves by enabling the ability to opt out of this behaviour.

istarkov:
I understand and this situation is by design, you need to track center position using `onBoundsChange` event and store `center` in flux store or in some kind of state like i do in this example 
[events_map_page](https://github.com/istarkov/google-map-react-examples/blob/master/web/flux/components/examples/x_events/events_map_page.jsx)

Other way is to add `defaultCenter` property to GoogleMap and i create [issue](https://github.com/istarkov/google-map-react/issues/10) on this and add as soon as have some time.
So in your example in the future you need just to define `defaultCenter` but for now use `onBoundsChange` event.

Chrisui:
Ah didn't see the onBoundsChange prop!

Much nicer thank you. :)

PR #40, opened 2015-09-29
Fix const name.

Andreyco:


istarkov:
My fault, no tests at all except this :-) https://github.com/istarkov/google-map-react/blob/master/src/__tests__/eye_test.js#L1

PR #58, opened 2015-10-20
support for international locales. 

martingg88:


istarkov:
Looks like we need to rewrite this to support other keys. 
For example I need soon:
key `client` https://developers.google.com/maps/documentation/business/clientside/#client_id
and key `libraries` https://developers.google.com/maps/documentation/javascript/libraries

So we need implementation like

```javascript
export default function googleMapLoader(api:string | object)
```

`string` to support previous behavior and `object` for a new

Where object keys is url query params.

so I can to load googleMap as before using api key, or use new `api` prop, with all keys I need
like 

```javascript
<GoogleMap api={{
  client: 'mysuperclient',
  libraries: 'a,b,c',
  ...etc
}}
```

If both new api object and old apiKey both presented we need to give a warning in dev mode and use api object.

istarkov:
May be rename `api` to other name something like `bootstrapURLParams` or something more good.

martingg88:
+1

istarkov:
Fixed here
https://github.com/istarkov/google-map-react/commit/862a6e85c45d6c9533e4dd9791969593a141cd91

Usage

```javascript
<GoogleMap 
  bootstrapURLKeys={{
    key: API_KEY,
    language: 'ru',
    ...otherUrlParams,
  }}
>
...
```

PR #74, opened 2015-12-04
Bugfix: _onGoogleApiLoaded() callback was being called on onZoom beca…

nopmop:
…use it was in the draw() method.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -396,6 +396,8 @@ export default class GoogleMap extends Component {
       this.map_ = map;
       this.maps_ = maps;
 
+      this._onGoogleApiLoaded({map, maps});
+
       // render in overlay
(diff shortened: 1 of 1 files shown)


istarkov:
Need some time to think, it was some Idea behind setting this event inside draw call.
As I remember this call in my idea must be after initialization step, because somebody can call setState in the callback, and setState inside initialization can cause undefined behavior.
I've just forgot that draw can be called on zoom event.

istarkov:
I think that we need to not move _onGoogleApiLoaded as I wrote above it can be dangerous, but just to add a guard `if(!_guardOnGoogleApiLoaded) { _onGoogleApiLoaded(); _guardOnGoogleApiLoaded = truse` ...`

PR #77, opened 2015-12-23
ensure calculation ends up with a positive integer

fidgety:
I've been using the utils to find the centre and zoom needed to show a number of markers on the visible map. Under some conditions we get no zoom level. I believe it's because we have some markers with negative latitude and some with positive. An Math.abs fixes the problem.

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -68,7 +68,7 @@ export default {
     const dx = nwWorld.x < seWorld.x
       ? seWorld.x - nwWorld.x
       : (1 - nwWorld.x) + seWorld.x;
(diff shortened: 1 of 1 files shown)


pezza3434:
This works perfectly for me @istarkov :+1:

fidgety:
My mistake - I was effectively passing in sw and ne. I'll close this PR and create a new one with a friendly error message if the parameters are invalid.

istarkov:
May be support both {nw, se} and {sw, ne}, as google use the second in LatLngBounds initialization?
Looks like my bounds variant is incorrect as all other apis uses {sw, ne} as bounds.

PR #114, opened 2016-03-14
update lodash version

mauriciosoares:
This PR is related to this issue I was having: https://github.com/istarkov/google-map-react/issues/113

Turned out this issue was specific to my building process (some weird bug), but I think that'd be good to update lodash version to 4.x

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -20,8 +20,8 @@ import pick from './utils/pick';
 import raf from './utils/raf';
 import log2 from './utils/math/log2';
 
-import assign from 'lodash/object/assign';
-import isNumber from 'lodash/lang/isNumber';
+import { assign } from 'lodash/object';
+import { isNumber } from 'lodash/lang';
 
 // To avoid Error with React 13, webpack will generate warning not error
 // more details is here https://github.com/orgsync/react-list/pull/54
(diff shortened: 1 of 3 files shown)


src/google_map.js
```diff
@@ -20,8 +20,8 @@ import pick from './utils/pick';
 import raf from './utils/raf';
 import log2 from './utils/math/log2';
 
-import assign from 'lodash/object/assign';
-import isNumber from 'lodash/lang/isNumber';
+import { assign } from 'lodash/object';
```
istarkov: just `import assign from 'lodash/assign'`

istarkov: otherwise it will add to build all lodash/object

istarkov: the same for others

mauriciosoares: yeah, that makes sense.

my bad

mauriciosoares:
@istarkov code updated.

istarkov:
Thank you!!!

mauriciosoares:
@istarkov could you please tell me when you publish this update?

I'll update my package.json as well :D

istarkov:
Wait for `npm install` :-) will finish.

istarkov:
google-map-react@0.10.6

mauriciosoares:
thanks!

PR #118, opened 2016-03-17
Added minZoomOverride attribute

ldong:
Added minZoomOverride attribute

If `minZoomOverride` is set to be true, options minZoom will be used instead of dynamically calculated. 

For example

```javascript
// Google Map Options
function createMapOptions(maps) {
    return {
        zoom: 2,
        minZoomOverride: true,
        minZoom: 2
    };
}
```

Please use with **caution**, as it can leads to [issues/83](https://github.com/istarkov/google-map-react/issues/83)

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -297,6 +297,10 @@ export default class GoogleMap extends Component {
   // otherwise there is no homeomorphism between screen coordinates and map
   // (one map coordinate can have different screen coordinates)
   _getMinZoom = () => {
+    if (this.options.minZoomOverride) {
+      return this.options.minZoom ? this.options.minZoom : DEFAULT_MIN_ZOOM;
+    }
+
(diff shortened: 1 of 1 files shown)


src/google_map.js
```diff
@@ -297,6 +297,10 @@ export default class GoogleMap extends Component {
   // otherwise there is no homeomorphism between screen coordinates and map
   // (one map coordinate can have different screen coordinates)
   _getMinZoom = () => {
+    if (this.options.minZoomOverride) {
```
istarkov: Thank you for your PR, but I think
this logic must be moved outside of this function, 
because with this logic it returns not a min possible zoom, but `minZoomOrSomethingFromOptions`. 

`this.options` must be removed, as it does not needed.

deviantony:
@ldong ping, could you update the PR?

istarkov:
Closed https://github.com/istarkov/google-map-react/pull/154

PR #121, opened 2016-03-21
Update README.md

stanleycyang:
Updated the docs for onChange

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ Example:
  <GoogleMap  onClick={_onClick} ... />
  ```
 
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you, as I know we are still have onBoundsChange method, and it's signature differs from onChange.

And onBoundsChange is deprecated, I think that it's better just to add 'deprecated' into documentation to onBoundsChange.

stanleycyang:
Added the deprecation warning :+1:

istarkov:
Thank you!!!!

PR #138, opened 2016-04-14
Remove es6-promise module

dapi:
Issue -  https://github.com/istarkov/google-map-react/issues/128

Was:

```
GoogleMapReact.min.js - 65732 bytes
```

Now:

```
GoogleMapReact.min.js - 58152 bytes
```

it's still works. All tests are passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "react": "^0.14.0 || ^15.0.0-rc.1"
   },
   "dependencies": {
-    "es6-promise": "^3.0.2",
     "eventemitter3": "^1.1.0",
     "lodash": "^4.6.1",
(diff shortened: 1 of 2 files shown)


istarkov:
Have you seen that tests ;-) (слезки а не тесты)

The problem is following:
I've added this promises library as there were a few issues about some browsers without promise support. At the beginning I wrote about polyfills - etc, but decided to add promise support for this.

IMHO If you think that promise should be removed (I think so) you need to add few lines into documentation how to add promise polyfill into user project.

dapi:
Really IE 11 still does not support Promise.
Well. I have added babel-polyfill. Now it is 147Kb in minified version )

PR #139, opened 2016-04-14
migrate from lodash to corejs functions

dapi:
Issue -  https://github.com/istarkov/google-map-react/issues/128

Was:

```
GoogleMapReact.min.js - 65732 bytes
```

Now:

```
GoogleMapReact.min.js - 46472 bytes
```

it's still works. All tests are passed.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -20,7 +20,6 @@ import pick from './utils/pick';
 import raf from './utils/raf';
 import log2 from './utils/math/log2';
 
-import assign from 'lodash/assign';
 import isNumber from 'lodash/isNumber';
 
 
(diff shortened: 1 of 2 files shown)


istarkov:
The same as here https://github.com/istarkov/google-map-react/pull/138
Initially library used corejs and move 2 lodash was because of users demand.

The question is - does any popular library asks users to use polyfills? 
I haven't seen that. It's good idea 2 move on corejs but it will be better 2 use `ponyfills` not `polyfills`.

This min build means not a lot, for example if I already use lodash this change does not affect my build size. But I use corejs ;-) so IMHO it's really better to switch on corejs but using `ponyfills`.

Call me on skype `cybice` if you have a questions or any issues with this.

PR #144, opened 2016-04-22
Add class name to marker

carsy:
It would be nice to have a class added to the marker to be able to animate its position. Even though animating top and left is not the most efficient method ever (transform is more performant) this is a good start (I can work on using transform to position the markers if you think it's a good idea).

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -247,7 +247,11 @@ export default class GoogleMapMarkers extends Component {
       };
 
       return (
-        <div key={childKey} style={{...style, ...stylePtPos}}>
+        <div
+          key={childKey}
+          style={{...style, ...stylePtPos}}
+          className={child.props.$markerHolderClassName}
+        >
           {React.cloneElement(child, {
             $hover: childKey === this.state.hoverKey,
             $getDimensions: this._getDimensions,
--- a/test/components/GoogleMap.spec.js
+++ b/test/components/GoogleMap.spec.js
@@ -175,4 +175,86 @@ describe('Components', () => {
       key: API_KEY,
     });
   });
+
+  it('Should add a className to the marker from $markerHolderClassName', () => {
+    const markerHolderClassName = 'marker-holder-class-name';
+
+    class MapHolder extends Component { // eslint-disable-line react/no-multi-comp
+      static propTypes = {
+        center: PropTypes.array,
+        zoom: PropTypes.number,
+        greatPlaceCoords: PropTypes.any,
+      };
+
+      static defaultProps = {
+        center: [59.938043, 30.337157],
+        zoom: 9,
+      };
+
+      constructor(props) {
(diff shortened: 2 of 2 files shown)


istarkov:
Im not prefer or use global css classnames at all, in all my current work I use css-modules. Also u can always use your marker animation, there is no need to animate component div, just animate your marker.

Also I could change internal lib details significantly, so really can't support this addition.

istarkov:
About left top - cant get your problem. Use transitions with offset on your marker

carsy:
@istarkov I completely agree with you. But ever you ever needed to animate markers in between coords? That's my use case.

istarkov:
Could you provide a real animation description, as I can animate anything independent of root element in most cases, so cant get your case :-(

istarkov:
Also now for a lot of animations I prefer to use react-motion like in cluster example (ref in docs),

carsy:
@istarkov sure thing, I'll get you an example tomorrow (plane trip).

istarkov:
Like here https://github.com/istarkov/google-map-clustering-example/blob/master/src/markers/ClusterMarker.js
For a lot of cases performance is really good, and allows to skip all that css hacks, like recreating elements to disable already running transitions etc. (Like 80% code of old markers is just to support css transitions)

carsy:
So my use case is to animate between two coordinates:

```jsx
<div
  lat={ this.props.position.lat }
  lng={ this.props.position.lng }
/>
```

Every time I call render, the lat and lng are different and I need to animate from the previous position to the new one.

istarkov:
You are right, In your case having className you could just set a transition in css. 
Ill write few moments later how to fix this PR.

carsy:
@istarkov thank you 👍

istarkov:
To make it best, let's pass className for this parent div on a marker, as it will be your class it will not pollute css. Also you could pass this className only for markers you need. (For example you want to animate just one marker)

Example
```javascript
<GMap>
  <MyMarker lat={lat} lng={lng} $markerHolderClassName={classNameForHolder} />
</GMap>
```

As here https://github.com/istarkov/google-map-react/pull/144/files#diff-7474dce7e0c5e0e592c272e34eefa16bR250 you have access to child props, you coud easily add this class. (be sure to not add anything if user has not pass this className)

Also for a time I want this `$markerHolderClassName` to be an `undocumented feature` with `$` char at beginning. (_I don't sure that it's a good idea to provide public access to this feature, until public example with this feature using will be provided_)

Thank you, will wait your PR.

carsy:
@istarkov on it ☺️

carsy:
@istarkov here you go

istarkov:
Cool! Thank you!!!

istarkov:
In npm as `0.12.0`

carsy:
@istarkov awesome 😄  thank you

PR #174, opened 2016-06-16
Added this.props.center check

stanislav-ermakov-roi:
I think there should be the check if old props center is set, otherwise latLng2Obj(this.props.center) fails.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -228,10 +228,14 @@ export default class GoogleMap extends Component {
 
     if (this.map_) {
       const centerLatLng = this.geoService_.getCenter();
-      if (nextProps.center) {
+      if (this._isCenterDefined(nextProps.center)) {
         const nextPropsCenter = latLng2Obj(nextProps.center);
-        const currCenter = latLng2Obj(this.props.center);
+        const currCenter = this._isCenterDefined(this.props.center) ?
+          latLng2Obj(this.props.center) :
+          null;
+
         if (
+          !currCenter ||
           Math.abs(nextPropsCenter.lat - currCenter.lat) +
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you, but code inside `if (` should work even if this.props.center is undefined, so the better is to fix 
something like this,
`const currCenter = this.props.center === undefined ? undefined : latLng2Obj(this.props.center);`

istarkov:
Good catch!

stanislav-ermakov-roi:
@istarkov, I guess if you do like this, code will fail at the next line:
`Math.abs(nextPropsCenter.lat - currCenter.lat)`
I think code inside `if (` is useless when we don't have `this.props.center` or it is `null`.

istarkov:
IMHO this should be run `this.map_.panTo({ lat: nextPropsCenter.lat, lng: nextPropsCenter.lng });` even if this.props.center is undefined

stanislav-ermakov-roi:
You are right, but probably not only undefined, but do a full check there using `this._isCenterDefined`, what do you think?

istarkov:
I think nothing ;-) (my coach bans me think) it's on you ;-)

stanislav-ermakov-roi:
:-D, please check new solution, just pushed.

istarkov:
eslint ` 233:1  error  Line 233 exceeds the maximum line length of 100  max-len`

istarkov:
All other looks good

stanislav-ermakov-roi:
Great, will run tests and lint locally to be sure this time :-)

istarkov:
Thank you, will publish today

istarkov:
google-map-react@0.14.5

PR #178, opened 2016-06-24
Calculate current touch position on click (fixes #28)

### LINKED ISSUE #28: Marker Taps Inconsistent (iOS)
Hi, it seems there is an issue with the `onChildClick` function on iOS, when tapping on a marker it doesn't register all the taps and rather than deselecting a selected marker and selecting a new one when tapping a new marker it seems to open and close the same balloon. I am experiencing this on the examples included here on all the browsers I've tested running iOS 8 on an iPhone 5 and 6+ (latest versions of Chrome, Safari, Opera mini). The problem exists on the iPad as well but to a lesser degree and becomes even less of a problem the larger the iPad is (iPad mini 1 is affected significantly also). Chrome on Android seems to work fine. Can anyone else confirm this?

BowlingX:
The current mouse/finger position is calculated on mousemove only right now. To support proper touch handling, a new position is now also calculated on click. This fixes #28.

There is another issue, if you move the finger around the map, for some reason the whole site moves too (iOS 9.3). So you can't navigate inside the map without scrolling the body. This is fixable with an onTouchMove (with `e.preventDefault()`) property on the parent div container of the map. I did not add it yet, because I'm not sure if this is a react issue or google did not prevent the events from bubbeling (what is weird because I can't reproduce this issue on a stand alone map without react).

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -725,6 +725,8 @@ export default class GoogleMap extends Component {
 
   _onMapClick = (event) => {
     if (this.markersDispatcher_) {
+      // support touch events and recalculate mouse position on click
+      this._onMapMouseMove(event);
       const currTime = (new Date()).getTime();
       if (currTime - this.dragTime_ > K_IDLE_TIMEOUT) {
         if (this.mouse_) {


istarkov:
Thank you, will publish today

istarkov:
In npm 0.15.0

BowlingX:
Thank you :)

PR #188, opened 2016-07-04
Fix wrong decription of fitBounds() function. 

oldwin:
`const {center, zoom} = fitBounds({nw, se}, size);`
{nw, se} are undefined

Need to pass bounds obect:
`const {center, zoom} = fitBounds(bounds, size);`


istarkov:
Thank you!!!

PR #197, opened 2016-07-18
Added hijack scroll function

alex-e-leon:
@istarkov I've added a hijack scroll function that allows you to manipulate the scroll event

I couldn't find any PR instructions so please let me know if there's any issues or if you'd like me for example to write tests.

Example uses:
* You'd like to zoom only multiples of 2 on scroll
* Implement a custom info-window (see https://github.com/istarkov/google-map-react/issues/180) or any other component that has a scroll bar. This is my use-case - I use this function to hijack the scroll event and then zoom only if I'm not inside a scrollable component.

Also note that I've added some example functions into the demo files to show how it works, but because the demo maps call `setMapParams` on `onChange` events you don't get a smooth zoom change but instead get a slightly slower zoom change from changing the zoom props.

I can remove these from the demo if you like but thought it might be useful for you to test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,9 @@ Example:
 
 #### googleMapLoader (func)
 
+#### hijackScroll(func)
+Allows you to takeover gmaps default scroll handler and manage scrolling over the map. Calls your function with the following arguments (scrollEvent, currentZoom, mapZoomFunction) where mapZoomFunction is map.setZoom(zoom) under the hood.
+
 #### onGoogleApiLoaded (func)
 Directly access the maps API - *use at your own risk!*
 
--- a/develop/GMap.js
+++ b/develop/GMap.js
@@ -15,7 +15,7 @@ import { susolvkaCoords, generateMarkers } from './data/fakeData';
 export const gMap = ({
   style, hoverDistance, options,
   mapParams: { center, zoom },
-  onChange, onChildMouseEnter, onChildMouseLeave,
+  onChange, onChildMouseEnter, onChildMouseLeave, hijackScroll,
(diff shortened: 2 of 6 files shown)


alex-e-leon:
@istarkov Any thoughts on this PR?

istarkov:
Thank you for contribution, IMO this must be done in the other not so hackable way.

> You'd like to zoom only multiples of 2 on scroll
Why? Im sure no one need this.

> Implement a custom info-window ...
It is the case, and it is almost the same case as for `draggable` map propery.

When I create draggable markers, I need to disable drag when I hover over the marker.
Here when I create info window with scroll, all I need is just disable scroll on the map.

I think all I need is just to allow `options` property to be dynamic (now options are readed at the start), so this be enough to solve your case.

istarkov:
PS: I'm on vacation now for next week or two, so I can answer fast only on simple issues and PRs as most time I use mobile phone for reading this.

alex-e-leon:
@istarkov Ah yeap, I agree, that's a better solution in some respects, though it has the implication of requiring one to move the logic to mouseEnter + mouseLeave (since otherwise by the time you try to catch the scroll event, it will have already propogated to gMaps so turning it off does nothing)

The frustrating thing about this, is that googleMaps has no support for mouseOver/mouseOut, just mouseEnter/mouseLeave which means these events trigger will trigger frequently as you move over a popup

alex-e-leon:
@istarkov is the dynamic options something you're currently working on already? Otherwise, I might have a go at implementing it and submitting a new PR.

istarkov:
@alex-e-leon It will be great if you make such PR. Thank you!

PR #204, opened 2016-07-24
Skip falsy markers

### LINKED ISSUE #203: Conditionally rendering marker
I have the following code, in which I am trying to conditionally render a marker:
```
        <GoogleMap>
        {
          userLocation
          ? <UserMarker key="user" {...userLocation} />
          : null
        }
        </GoogleMap>
```

However I'm getting the following error:
```
Uncaught TypeError: Cannot read property 'props' of null
google_map_markers.js:244
```

The code triggering the error seems to be this:

image

As react allows passing null children I think it would be appropriate to check if the child is null.

shark0der:
Fixes #203

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -229,6 +229,7 @@ export default class GoogleMapMarkers extends Component {
     this.dimesionsCache_ = {};
 
     const markers = React.Children.map(this.state.children, (child, childIndex) => {
+      if (!child) return undefined;
       if (child.props.latLng === undefined &&
           child.props.lat === undefined &&
           child.props.lng === undefined) {


src/google_map_markers.js
```diff
@@ -229,6 +229,7 @@ export default class GoogleMapMarkers extends Component {
     this.dimesionsCache_ = {};
 
     const markers = React.Children.map(this.state.children, (child, childIndex) => {
+      if (!child) return null;
```
istarkov: why null? why not undefined or why not a filter like util

shark0der: I've commented on the issue as well:

> I've returned null if child is falsy. The resulting array is passed to react so no need to filter out them after map.

dapetcu21: Also, I can see this library being used with big amounts of markers, so minimising stepts of intermediate array construction (with `.map` and `.filter`), might be a good thing.

istarkov: I don't like `null`s ;-) the same result we will get with `undefined`. 
So the question was about does `null` here has any advantage or hidden sense over `undefined`. 

Having ES6 using `null` in any chain like env prevents me to use arguments defaults in the future, so I prefer not to use `null` at all. 

So please change null on undefined ;-)

shark0der:
I've changed `null` to `undefined` as per request, however, you should keep in mind that null is generally more acceptable than undefined.

References:

Early react didn't even have proper validation for undefined:
https://github.com/facebook/react/issues/1647 

> Therefore passing undefined may actually change behavior of underlying classes in confusing ways. 

https://github.com/facebook/react/pull/1668#issuecomment-52244027

istarkov:
Thank you!
(IMO both `null` and `undefined` should gone from all languages ;-))

istarkov:
npm 0.16.2

PR #211, opened 2016-08-02
call zoomControlClickTime on all clicks

alex-e-leon:
fixes an issue where zoomControlClickTime doesn't get called on any clicks not handled by google maps, so if a marker click function should change the zoom level you receive the `GoogleMap bounds not eq:` error.


jooj123:
Is this because `onMouseMove` event not called on touch causing `mouse_` to be null?

istarkov:
Thank you!,
Does it visually changes something? The problem was that in chrome zoom had some bad visual effects, and this partially solved that behavior. 
Now I don't remember the idea why I move that code line into `if` but think that there were some idea, will check in few days.
Also looks like `GoogleMap bounds not eq:` warning should be deleted (or guarded with some `if`) as in most cases this messages are useless.

alex-e-leon:
this seems to look ok in chrome.

istarkov:
I need to check this with some of our production code, which causes an addition of double raf hack, it will take some time.

itsmichaeldiego:
@alex-e-leon Thanks for your contribution. We would like to have this change but it is having conflicts. Would you mind to add a commit to fix it? Thanks.

alex-e-leon:
@itsmichaeldiego updated

itsmichaeldiego:
@alex-e-leon Thanks, I am still a little bit worried about what @istarkov mentioned [earlier](https://github.com/istarkov/google-map-react/pull/211#issuecomment-236817326). As this fix was supposed to improve chrome's visual behavior. Would you mind sharing a test-video? Otherwise give me some time and I will test it thoroughly. Thanks in advance.

alex-e-leon:
No problem @itsmichaeldiego I ended up working around the bug, and I haven't worked on this problem in a while, so I don't have any video I can share. I also don't mind if you close it tbh, but if you feel like working on it let me know how you go!

itsmichaeldiego:
@alex-e-leon Oh Alex, thanks and no worries. I will take a look and test it as soon as I have some spare time. Thanks for fixing the conflicts!

itsmichaeldiego:
@alex-e-leon I will merge this soon for next release.

itsmichaeldiego:
@alex-e-leon I am going to merge this soon, sorry for the delay!

PR #233, opened 2016-09-01
Fix ipad dragging issue

jooj123:
Fixes https://github.com/istarkov/google-map-react/issues/232

Before: 
before

After: 
after

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -820,7 +820,12 @@ export default class GoogleMap extends Component {
   }
 
   _onTouchMove = (event) => {
-    event.preventDefault();
+    if (this.refs.google_map_dom) {
+      const mapDom = ReactDOM.findDOMNode(this.refs.google_map_dom);
+      if (mapDom.contains(event.target)) {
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you!
So map somethere in the page will prevent the whole page from moving as usual?
Also this events could be added on the user side? May be just extend help with this solution?

jooj123:
I have added this at my end for now - and that works as well.
 
But it should be in this component because you wont ever need to move the entire page as you drag the map - what would be the purpose of that?

istarkov:
For example we have a page:
```
<content height={100vh}/>
<content height={100vh}/>
<map/>
```
Map is on the third screen from top, but prevents the whole page touchmove events,
it's bad when library prevents something in the world to happen.

jooj123:
Oh i see what your saying - let me fix the PR ..

istarkov:
Super!!!

istarkov:
Thank you! 
in npm 0.19.0

PR #241, opened 2016-09-16
doc: clarify `onClick` and `onChange` arg structure

kevmannn:
More clearly show that `onClick` and `onChange` take objects as arguments.


istarkov:
Thank you!!!

PR #246, opened 2016-09-22
Fix unreliable `window` and `document` detection

viktorbezdek:
The method used for detecting DOM availability was sloppy and breaks server side rendering in some cases. This new approach fix cases when global variable window is defined in NodeJS by checking presence of `window.document` and `window.document.createElement` which are used in element resize detection.


istarkov:
1) Please remove all .gitignore files, https://github.com/istarkov/google-map-react/blob/master/.gitignore does your git tool not support that?
2) `was sloppy and breaks server side rendering in some cases` - please provide that cases and why just typeof window is not enough.

viktorbezdek:
1) Dammit, I committed compiled files to my fork just to be able to use my fork as a dependency temporarily before you publish the patch. Should have make topic branch for the pull request.
2) One of the cases is quite simple - my application have to use private module that I have no control of. This module requires global variable window with some unrelated data, so `typeof window !== 'undefined'` become truthy and SSR breaks. New solution only makes deeper more robust checking. There is another, more tricky, case but I have to debug my app I bit more to make sure it's not issue on my side.

I'll fix files in the repo in sec. Will let you know.

viktorbezdek:
The last commit is reverted. If you accept the pull request, please publish the patch asap.

istarkov:
Thank you!!! Will publish today.

PR #249, opened 2016-10-01
Compare LatLngLiteral objects by value to avoid warnings

NoxHarmonium:
The way my web application is set up, it uses a lot of immutable objects that are copied. This means that the properties I feed into the GoogleMap component can be equal according to their values, but not by reference because they are technically different objects.

I was getting a lot of warnings from google-map-react saying that I was changing the `defaultCenter` property after it was initially set. I checked and the values were not actually changing but each object was new so the `===` equality comparison was not working in that instance.

I have added a function to compare by value. Do you think this is a worthwhile change? I have tested it locally in my project and it seems to be working well.

I didn't know whether to use a shallow compare library like `react-addons-shallow-compare` (https://facebook.github.io/react/docs/shallow-compare.html). On one hand a 3rd party library could possibly cover edge cases I hadn't thought of but also it is another dependency to pull into the project and make the final file size bigger.

Let me know what you think.

Thanks!

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -23,6 +23,7 @@ import log2 from './utils/math/log2';
 import isNumber from './utils/isNumber';
 import omit from './utils/omit';
 import detectElementResize from './utils/detectElementResize';
+import latLngAreEqual from './utils/latLngAreEqual';
 
 const kEPS = 0.00001;
 const K_GOOGLE_TILE_SIZE = 256;
@@ -220,7 +221,7 @@ export default class GoogleMap extends Component {
 
   componentWillReceiveProps(nextProps) {
     if (process.env.NODE_ENV !== 'production') {
-      if (this.props.defaultCenter !== nextProps.defaultCenter) {
+      if (!latLngAreEqual(this.props.defaultCenter, nextProps.defaultCenter)) {
         console.warn('GoogleMap: defaultCenter prop changed. ' +  // eslint-disable-line
                       'You can\'t change default props.');
       }
--- a/src/utils/latLngAreEqual.js
+++ b/src/utils/latLngAreEqual.js
@@ -1,0 +1,15 @@
+// Determines if two LatLngLiteral objects are equal, not just by reference
+// but by their properties
+export default function latLngAreEqual(latLng1, latLng2) {
+  if (latLng1 === latLng2) {
+    // Reference equality
+    return true;
+  }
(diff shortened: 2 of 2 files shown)


istarkov:
In almost all cases I would prefer reference equality check on any component prop. 
IMO conception of immutability does not mean that objects are copied on any operation, but that any unchanged object will have the same reference.
Such idea is used almost everythere for react components optimization. And I think that any custom equality checks on props should not be used, just shallowEqual.
So this warning was set to show that something is wrong with the data. As it will cause unneeded component rendering.
Also in most cases you does not need defaultBlabla props,  I want to remove them in near future :-)

istarkov:
BTW I see the case where defaultBlabla props could be different, for example carousel cards with map, so the warning you have will occure on map reuse. So it could occure even data is not wrong.
IMO is better to not use defaultBlbla props as now Im really sure that I should delete them from API

NoxHarmonium:
Yeah actually I believe you're right, pure react components do a shallow comparison between previous props/state and current props/state to determine of they need updating. Therefore, since `defaultCenter` is a property it should be reference compared.

I have a custom implementation of pure components that performs a deep equality comparison for certain types of objects but it is not the standard react way of doing things and I shouldn't apply it to this library.

I should have thought about it more before sending the PR. 

Would setting `defaultCenter` to some arbitrary value fix the warning but not actually affect the app because it will be overridden immediately with the `center` property.

istarkov:
In most cases I do not use defaultCenter property at all in my current apps,
like [here](https://github.com/istarkov/google-map-clustering-example/blob/master/src/GMap.js#L19-L27)

That parameters is needed for using map in [uncontrollable](https://facebook.github.io/react/docs/forms.html#controlled-components) way, and now I see no reason to use any component as uncontrolled.
Where is no need IMO in default props for controlled components at all.
Even you have undefined center prop you could always define it via js in multiple ways `(center || default)`  or set default at function arguments of functional react component `({ center = DEFAULT_CENTER }) =>` or set via recompose `defaultProps` enhancer.

So I just recommend to not use it at all.
 
Having my experience every custom equality could be a problem in the future as other developers in your team could be confused trying to find why some component does not updated when it should. 
Few times in the past a big amount of time was wasted for investigation process of strange update behaviour.

NoxHarmonium:
Thanks for the tip!

PR #250, opened 2016-10-02
Prevent new Promise() call before initializing polyfill.

arkist:
Prevent `new Promise()` call before initializing polyfill.

Currently, it calls `new Promise()` in outside of the function. 
if you bundle this component with entry module via Webpack, `Promise is not defined` error occurs on page initialize. IE and many android devices still have this problem.

### DIFF
--- a/src/utils/loaders/google_map_loader.js
+++ b/src/utils/loaders/google_map_loader.js
@@ -3,12 +3,13 @@ let $script_ = null;
 let loadPromise_;
 
 let resolveCustomPromise_;
-const _customPromise = new Promise(resolve => {
-  resolveCustomPromise_ = resolve;
-});
 
 // TODO add libraries language and other map options
 export default function googleMapLoader(bootstrapURLKeys) {
+  const _customPromise = new Promise(resolve => {
+    resolveCustomPromise_ = resolve;
+  });
+
   if (!$script_) {
     $script_ = require('scriptjs'); // eslint-disable-line
   }


istarkov:
Hi thank you, 
the patch is not the same I logic I used, 
every new call to loader new Promise will be created so no any previous promises will be resolved.
And this logic is not the logic I need. 

Also please describe how it is possible that polyfill is not exists on the moment of  Promise is used.  Why it will exists at the moment the loader called? For example in some cases I call the loader not from google map but from the code, this will not solve this.

I think if you support older browsers you just need bundle a piece of your code, not the library itself, and just include polyfill there.

arkist:
@istarkov Thanks for quick response! :)
I missed about logic. it's weird but this problem is a real. if you run the demo (`npm run start`) using IE, you can see the same error.

istarkov:
As I understand there is no polyfills included in demo ;-)

arkist:
Yes, I tested with es6-polyfill in demo's entry. miss that thing so I remove comment but you already checked 😂 I'll check another way.

istarkov:
May be just add a readable warning about that polyfill should be included before map is used if `typeof Promise === 'undefined' && process.env.NODE_ENV==='DEBUG'` as 
even your move `Promise` usage inside loader it will not give a guarantee that Promise polyfill is already loaded in other bundle.

arkist:
Thanks for the kind advice!
I used `require('es6-promise').polyfill()` on my entry, but now I just throw away that thing and use babel-polyfill. 
it's not the correct answer to this issue but is works. 😇

PR #272, opened 2016-11-03
#add handlers for fly update layers

ipostol:
Hello, this is update for resolve fly updates layerTypes by new props (now they can inititialize with map 1 time)


src/google_map.js
```diff
@@ -291,6 +294,14 @@ export default class GoogleMap extends Component {
 
         this.map_.setOptions(options);
       }
+
+      if (!isEqual(nextProps.layerTypes.sort(), this.props.layerTypes.sort())) {
```
istarkov: imo it's on user to sort layers, and to support referential equality. I don't like any kind of isEqual + sort checks here.

ipostol:
@istarkov, remove sort/isEqual logic, left only ref equal

istarkov:
Thank you!

istarkov:
0.21.3

PR #273, opened 2016-11-07
Fixed componentWillReceiveProps options function bug

alex-e-leon:
Fixes a bug which prevented you from changing the function passed to the options prop


istarkov:
Super!!!
Thank you!

istarkov:
google-map-react@0.21.2

PR #275, opened 2016-11-07
expose map type callback

mattydoincode:
This PR would expose the google maps event that says "hey, the user changed from satellite view to roads view" or vice versa. This is something I need to detect and change the icon styles based on so am looking to expose it. Any input is greatly appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,9 @@ Example:
 
 #### onZoomAnimationEnd (func)
 
+#### onMapTypeIdChange (func)
+When the user changes the map type (HYBRID, ROADMAP, SATELLITE, TERRAIN) this fires
+
 #### distanceToMouse (func)
 
 #### googleMapLoader (func)
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -80,6 +80,7 @@ export default class GoogleMap extends Component {
     onZoomAnimationStart: PropTypes.func,
     onZoomAnimationEnd: PropTypes.func,
     onDrag: PropTypes.func,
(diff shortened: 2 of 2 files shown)


mattydoincode:
Doesn't seem like CI is related to my PR...

istarkov:
Really?
653:63  error  Trailing spaces not allowed  no-trailing-spaces

mattydoincode:
LOL when i looked at it that message hadn't loaded yet and I got super confused. Will update!!

istarkov:
Having `onBlabblaChange` calback in interface I always try to find `blabbla` prop on any Component public interface. 
Here we have 'onMapTypeIdChange' but `mapTypeId` is not present in props, for me it's confusing a little.
Not sure what to do, or to add `mapTypeId` as a controllable property, or to somehow change the event name. IMO first is overhead, and I can't suggest a good name for the second.
Your opinion?

mattydoincode:
That's a great point, I would say considering it's something that a user might want to set.. for example "oh, i want to show a satellite map here" we probably want it to be controlled. This seems reasonable to me.
https://developers.google.com/maps/documentation/javascript/reference#MapTypeId. The real truth though, is isn't it wrapped up in the options object? Like a lot of the things you can do with this library are hidden behind that sort of big options object that you can fill with styles, or whatever you want, so in a sense it's kinda handled. Thoughts?

istarkov:
Please add 1-2 lines about in documentation, and let's merge then.
 
Let's think about after merge, as now I have no ideas ;-)

istarkov:
google-map-react@0.21.4

PR #278, opened 2016-11-17
Geojson

mattydoincode:
this adds dope features!

* geojson layers
* map type callback


mattydoincode:
Meant to merge this locally, I apologize!

PR #282, opened 2016-11-24
Keep original center on _mapDomResizeCallback

Andries-Smit:
When the resize event is triggered, the original center was lost. 
This issue is significant when the map is renders outside visible dom with zero height and width.
When placed into the visible dom, the resize is triggered correctly. With this change the center will persist. (Without this fix the old center is the top left corner)

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -390,7 +390,9 @@ export default class GoogleMap extends Component {
   _mapDomResizeCallback = () => {
     this.resetSizeOnIdle_ = true;
     if (this.maps_) {
+      const originalCenter = this.maps_.getCenter();
       this.maps_.event.trigger(this.map_, 'resize');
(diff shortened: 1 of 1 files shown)


istarkov:
Any picture? Or jsbin example please with the problem

Andries-Smit:
https://jsbin.com/yevuzaboba/edit?js,console,output

istarkov:
I see, thank you!!!

istarkov:
google-map-react@0.21.4

Andries-Smit:
Thanks for the quick response, though I made a small mistake. I opened a new pull request. #283

gablabelle:
@istarkov I tried to use the `lockCenter` prop without success.

How can I make sure that if I resize the map's container the center (and my marker) will stay in ... wait for it ... the center? ;-)

PR #283, opened 2016-11-24
Fix reference to maps instead of map

Andries-Smit:
Sorry, made a smal but ugly mistake in the last pull request.
The this.maps does not have a function getCenter .... but this.map does.


Andries-Smit:
This was the result with 0.21.4 https://jsbin.com/yevuzaboba/edit?html,js,console,output

mistaguy:
Confirmed, this fixes the resize on my side

istarkov:
`npm run start` to run dev env to avoid errors ;-)

PR #284, opened 2016-11-24
Clustering and Spiderify

mukulb90:
We use Google-map-react very heavily in our projects and like many others, we too ran into the problem of too many markers. Most obvious solution to solve it was Clustering and Spiderify. 

Clustering ->  Two or more markers join together to form a cluster .

Spiderify -> When two or more markers are at exactly same latLng, plotting a lot of markers gave a very bad UX. We had a use case where we had to plot more than 100 markers at exactly same latLng .

I read a solution where you have mentioned to use point-cluster to form clusters externally and pass clusters as normal markers to Google-Map-React . 

On the same lines, I have Implemented an HOC : GoogleMapReactWithClustering
Which enhances GoogleMapReact with clustering and spiderifying where we just need to pass React Components representing Cluster and Spiders and it will do the rest.
Also, one of the requirements for us was that clusters should split smoothly so that you can make sense of which marker came from which cluster on zooming in.

This is how it works for us :- http://jmp.sh/zeGtADw

I wasn't sure what directory structure to follow . So, i have put src code in develop/clustering for now .


istarkov:
Super!
IMO it will be `10x` better if you provide this as your own standalone enhancer library and provide a Pull Request with link to your library in current Readme.

Tim152:
@istarkov hi! I was create a simple example without recompose, based on your [desicion](https://github.com/istarkov/google-map-clustering-example) 
https://github.com/Tim152/clustering-google-map-react

itsmichaeldiego:
@Tim152 Feel free to do what @istarkov said before and add the enhacer to our Readme

itsmichaeldiego:
@mukulb90 You still active?

mukulb90:
Yes michael

itsmichaeldiego:
@mukulb90 I would like to work with you to see if we can add this feature

mukulb90:
Sure, What do you have in mind ?

itsmichaeldiego:
@mukulb90 Well, first, we would have to fix the conflicts and then we can go two ways, we've a a organization now, so we can have a "standalone enhancer library" as @istarkov pointed out, or we can discuss if we want to add clustering to this one. 

I would like to see both options but first I would need you to fix the conflicts so I can test it, and test the size of the build.

Would you mind fixing conflicts and updating to latest master?

Thanks! 🎉

mukulb90:
@itsmichaeldiego I have fixed the conflicts here

itsmichaeldiego:
@mukulb90 Did you check? I just did a find in page and found it https://github.com/google-map-react/google-map-react/pull/284/files#diff-e8b86504901df43bfb0218a9db49bb7bR95

taylorjames:
hey @itsmichaeldiego what is the status of this? I would like to have clustering in my implementation.

natcodes:
hello @itsmichaeldiego @mukulb90 what is the status of this? I would also like to use clustering.

GodLesZ:
@mukulb90 @itsmichaeldiego why was this issue closed?
Any news on for that feature?

itsmichaeldiego:
@taylorjames @GodLesZ @natcodes My ideal idea was to create another library because this changes might affect this one. But if anybody wants to create a PR similar to @mukulb90's, I am happy to help review and merge!

PR #287, opened 2016-11-28
Add ne and sw to bounds object

lukyanovfedor:
Hi, thanks for the great library it serving to me faithfully not first time. 

I found small problem with bounds object, most geo tools expect to receive data with NE and SW, instead of NW and SE. So i thought it will be cool if bounds object will contain all corners.

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -116,21 +116,22 @@ function fitNwSe(nw, se, width, height) {
 
 const exports = {
   fitBounds({ nw, se, ne, sw }, { width, height }) {
+    let fittedData;
+
     if (nw && se) {
-      return fitNwSe(nw, se, width, height);
+      fittedData = fitNwSe(nw, se, width, height);
+    } else {
+      const calculatedNwSe = this.convertNeSwToNwSe({ ne, sw });
+      fittedData = fitNwSe(calculatedNwSe.nw, calculatedNwSe.se, width, height);
     }
 
-    const calculatedNw = {
-      lat: ne.lat,
-      lng: sw.lng,
+    return {
+      ...fittedData,
+      newBounds: {
+        ...fittedData.newBounds,
+        ...this.convertNwSeToNeSw(fittedData.newBounds),
+      },
     };
-
-    const calculatedSe = {
-      lat: sw.lat,
-      lng: ne.lng,
-    };
-
-    return fitNwSe(calculatedNw, calculatedSe, width, height);
   },
 
   // -------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


istarkov:
Hi thank you, fully agree. Will check PR later today or tomorrow.

istarkov:
All is fine, may be add ability to use NE SW (if NW SE is not provided) to documented [fitBounds](https://github.com/istarkov/google-map-react/blob/7ae63cb3d733b91ef14332ed52dd38b5886459f5/src/utils/utils.js#L67) ?

lukyanovfedor:
@istarkov It's make sense, what do you think about such approach: 

1) If we have NW and SE we're just using them.
2) If we have NE and SW we're converting them into NW and SE and using same algorithm that we have in `fitBounds` method.

istarkov:
Super!

lukyanovfedor:
@istarkov cool, i'll push changes tomorrow morning

istarkov:
Thank you!

lukyanovfedor:
@istarkov hm, I thought a little bit about it today, and i don't really like that we're passing NE and SW to `fitBounds`, but it returns NW and SE. What do you think about it? 

Also i think will can add some helpers like `convertNeSwToNwSe` and `convertNwSeToNeSw`, cause we're repeating this logic in few places.

istarkov:
Agree that newBounds must have all corners,
any your decision on `convertXXX2YYY` will be good.

lukyanovfedor:
@istarkov done!

src/utils/utils.js
```diff
@@ -63,54 +63,73 @@ function meters2WorldSize(meters, { lat, lng }) {
...
+    center: world2LatLng(middle),
+    zoom,
+    newBounds: {
+      nw: newNW,
+      se: newSE,
+    },
+  };
+}
+
+const exports = {
```
istarkov: May be we will remove exports and all that `this` below, as it was time when `export default { blabla }` was the same as `export const blabla`
Having that it was deprecated in babel there is no need in exports object at all.

lukyanovfedor: Oh, that part seemed strange to me) Sure.

lukyanovfedor: Done!

istarkov:
Super, 
from first days I had knowledge that bounds object is not similar to others API 
(and my laziness didn't allow me to fix that)

all that my
```
  const ne = { lat: nw.lat, lng: se.lng };
  const sw = { lat: se.lat, lng: nw.lng };
  return new LatLngBounds(sw, ne);
```
now will gone ;-)

Thank you a lot!!!!

istarkov:
At npm google-map-react@0.22.0

lukyanovfedor:
Cool, glad that i helped 😉

PR #291, opened 2016-12-14
fix bug with inferno

mariusandra:
Hey, I'm trying to migrate from React to InfernoJS. Everything works great, except somehow in `componentWillUnmount` the DOM is already gone (`mapDom === null`) and this line throws an exception.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "google-map-react",
-  "version": "0.21.5",
+  "version": "0.22.0",
   "description": "isomorphic google map react component, allows render react components on the google map",
   "main": "lib/index",
   "scripts": {


mariusandra:
I think there was a typo there as well. `addResizeListener` should be `removeResizeListener`

src/google_map.js
```diff
@@ -328,10 +328,12 @@ export default class GoogleMap extends Component {
     const mapDom = ReactDOM.findDOMNode(this.refs.google_map_dom);
     window.removeEventListener('resize', this._onWindowResize);
     window.removeEventListener('keydown', this._onKeyDownCapture);
-    mapDom.removeEventListener('mousedown', this._onMapMouseDownNative, true);
```
istarkov: Why findDOMNode returns `null` on Inferno?

src/google_map.js
```diff
@@ -328,10 +328,12 @@ export default class GoogleMap extends Component {
...
     window.removeEventListener('mouseup', this._onChildMouseUp, false);
     window.removeEventListener('touchmove', this._onTouchMove);
-    detectElementResize.addResizeListener(mapDom, that._mapDomResizeCallback);
```
istarkov: Yes looks like mistake

src/google_map.js
mariusandra: Not sure. Seems like a mistake with inferno. I reported it here: https://github.com/trueadm/inferno/issues/590 . Not sure if/when it will be fixed.

istarkov:
Ill merge today evening or tomorrow morning, its fun to see that libraries work with inferno preact etc, wanna to play with too

src/google_map.js
```diff
@@ -328,10 +328,12 @@ export default class GoogleMap extends Component {
...
+    if (mapDom) {
```
istarkov: could u add a comment that this check is for inferno support

PR #295, opened 2016-12-22
fix small typo (add -> remove)

mariusandra:
Hey, my previous PR #291 is no longer needed. They fixed the bug in Inferno in the mean while! 

https://github.com/trueadm/inferno/issues/590#issuecomment-268852009

Here's a new PR to just fix that small add/remove typo.


istarkov:
Thank you!

Issue #300, opened 2017-01-10
Readme references MyGreatPlace, without an example.


aaronmader:
From the readme, section: [What's it Look Like?](https://github.com/istarkov/google-map-react#whats-it-look-like) 

    import MyGreatPlace from './my_great_place.jsx';

But this file doesn't exist within the repository. Perhaps the readme/examples could include the declaration of this expression, so new users aren't stranded with an example that can't compile?

rob-blasetti:
+1

alan345:
+1

solancer:
+1

sudhirbitsgoa:
+1

itsmichaeldiego:
@aaronmader @rob-blasetti @alan345 @solancer @sudhirbitsgoa This issue is outdated, Readme has been improved, please feel free to share your feedback, thanks!

PR #304, opened 2017-01-16
Improve the example code

aaronmader:
Update the example code to include a marker definition (MyGreatPlace) for completeness.
Addresses issue #300

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,10 +55,39 @@ In the simple case you just need to add `lat` `lng` props to any child of `Googl
 
 ```javascript
 import React, {PropTypes, Component} from 'react';
-import shouldPureComponentUpdate from 'react-pure-render/function';
-
 import GoogleMap from 'google-map-react';
-import MyGreatPlace from './my_great_place.jsx';
+
+const K_WIDTH = 40;
+const K_HEIGHT = 40;
+
+const greatPlaceStyle = {
+  position: 'absolute',
+  width: K_WIDTH,
+  height: K_HEIGHT,
+  // initially any map object has left top corner at lat lng coordinates
+  // it's on you to set object origin to 0,0 coordinates
+  left: -K_WIDTH / 2,
+  top: -K_HEIGHT / 2,
+
+  border: '5px solid #f44336',
+  borderRadius: K_HEIGHT,
+  backgroundColor: 'white',
+  textAlign: 'center',
+  color: '#3f51b5',
+  fontSize: 16,
+  fontWeight: 'bold',
+  padding: 4
+};
+
+class MyGreatPlace extends Component {
+  render() {
+    return (
+       <div style={greatPlaceStyle}>
+          {this.props.text}
+       </div>
+    );
+  }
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you! But -

This is readme example. 
- It must be short. 
- Just before example I wrote `you just need to add lat lng props to any child of GoogleMap component`
- At first sentences of readme I wrote `It allows you to render any React component on the Google Map`
- `import` contains `./` which means that it's some your component
- Where are a lot of working examples refs in readme even on `jsbin`.

So sorry I don't think this is needed.

aaronmader:
Solid points.. I did eventually find my way over to https://github.com/istarkov/google-map-react-examples which was a HUGE help. 

I got stuck because I was assuming that there was more to the MyGreatPlace component. (And I suppose that's where the other +1's in the issue were struggling too). You're instruction `you just need to add lat lng props to any child of GoogleMap component` makes sense to me *now*, but at the time I was just trying to copy/paste your example into my react project, without success. 

Maybe if we just add another simple example using a div? (I'll create a pull request to show what I mean)

Sapphire64:
I was thinking about improving that part too and suddenly found this PR already in place :) `MyGreatPlace` has to be explained, for me passing of geo attributes to every kinds of components seems to be unclear (but I agree, this is useful), so I had to dig into examples too. I thought you need to handle the coordinates inside of the component.

So I am +1 on current patch or maybe can suggest to apply following patch to have only one usage of "MyGreatPlace" in the example:
```
- <MyGreatPlace lat={59.955413} lng={30.337844} text={'A'} /* Kreyser Avrora */ />
+ <div lat={59.955413} lng={30.337844}>{'A'}</div> /* Kreyser Avrora */
```

`MyGreatPlace` is also vague to be fair, if we change it to something in the context like `RedMarker` it will explain more that it's a wrapper, not the actual target object with coordinates like you can misunderstand now.

In general I would prefer to have next interface when it comes to markers on the map to make it less confusing:
```
<Marker key={} lat={} lng={}>
    <YourComponentHere/>
</Marker>
```
But that's another story.

hjmccain:
The lack of detail about MyGreatPlace was a source of frustration for me, as well, and led me to initially pass up this module. @istarkov your argument is that the Readme needs to be short — explaining MyGreatPlace module would barely lengthen it at all — four sentences, maybe, and a line or two of code? Worth it for the clarity!

hjmccain:
@aaronmader I'm looking at the pull request you made, but when I try to add a div with lat & lng, the marker is not showing up and I'm getting some complaining from the console — `Warning: Unknown props `lat`, `lng`, `$hover`, `$getDimensions`, `$dimensionKey`, `$geoService`, `$onMouseAllow`, `$prerender` on <div> tag.` If you're around and down to help clarify your code — I'm probably missing something small — I'd really appreciate it!

istarkov:
Thank you all, please look at current readme
https://github.com/istarkov/google-map-react/pull/320

PR #308, opened 2017-01-23
Lock / retain specified center on resize

Marcel-G:
#282 

Some way to keep the map centered on center / default center prop when _mapDomResizeCallback is called would be useful.

I'm using google-map-block with panControl and dragging disabled, therefore the center cannot change from user input. When the map is resized, I want the map to center on the center prop given to google-map-react not getCenter()

```
<GoogleMap
  options={{
    panControl: false,
    draggable: false
  }}
  center={{lat: 59.938043, lng: 30.337157}}
  zoom={9}
  lockCenter>
    {Markers}
</GoogleMap>
```


src/google_map.js
```diff
@@ -453,9 +454,10 @@ export default class GoogleMap extends Component {
   _mapDomResizeCallback = () => {
     this.resetSizeOnIdle_ = true;
     if (this.maps_) {
-      const originalCenter = this.map_.getCenter();
+      const originalCenter = this.props.center || this.props.defaultCenter;
+      const currentCenter = this.map_.getCenter();
       this.maps_.event.trigger(this.map_, 'resize');
-      this.map_.setCenter(originalCenter);
+      this.map_.setCenter(this.props.lockCenter ? originalCenter : currentCenter);
```
itsmichaeldiego: @Marcel-G This PR is looking good! But it is throwing an error in this line on Travis CI:

`460:27  error  Follow `prettier` formatting (expected '\n' but found 't')  prettier/prettier`

Would you mind checking on that?

josephfrazier: @itsmichaeldiego, thanks for stepping up as a new maintainer! I've already upgraded to the latest version, thanks to the fix in https://github.com/istarkov/google-map-react/pull/452#issuecomment-353203783. However, I need to get this fix landed as well before I can actually take advantage of the fullscreen functionality.

The CI error can be fixed by running `yarn lint --fix`, or applying the following patch:

```diff
diff --git a/src/google_map.js b/src/google_map.js
index aba1ea5..e4a8b8c 100644
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -457,7 +457,9 @@ export default class GoogleMap extends Component {
       const originalCenter = this.props.center || this.props.defaultCenter;
       const currentCenter = this.map_.getCenter();
       this.maps_.event.trigger(this.map_, 'resize');
-      this.map_.setCenter(this.props.lockCenter ? originalCenter : currentCenter);
+      this.map_.setCenter(
+        this.props.lockCenter ? originalCenter : currentCenter
+      );
     }
   };
 
```

Would you like me to submit a separate PR?

itsmichaeldiego: @josephfrazier Thanks for the kind words. This fix seems like the solution to our problems. Feel free to make a PR, that would be awesome! We would also need a modification in the docs so users know they're able to use `lockCenter`

josephfrazier: Thanks for the quick feedback, see https://github.com/istarkov/google-map-react/pull/482

viperfx07:
Hi 
_mapDomResizeCallback is not triggered when I press the fullscreen button on the map.
I've set **resetBoundsOnResize** to true

test url:
http://msnlhd.surge.sh/detail

tested
on Chrome  v65.0.3325.18, Firefox 59.0.2 and IE11

I'm using google-map-react v0.34.0

Issue #297, opened 2017-01-06
Maps cooperative gesture handling


zeeke:
Hi,

The cooperative gesture handling described at [this link in the documentation](https://developers.google.com/maps/documentation/javascript/interaction) is not working due to the iPad fix in #233

```js
  // src/google_map.js:867

 _onTouchMove = (event) => {
    if (this.refs.google_map_dom) {
      const mapDom = ReactDOM.findDOMNode(this.refs.google_map_dom);
      if (mapDom.contains(event.target)) {
        event.preventDefault();
      }
    }
  }	
```

I don't know how it would be on iPad without this fix. At the moment, scolling with a single finger on the map makes a warning appear but the page does not scroll. With a big map there is a chance that the user fall in a section where he can't scroll away.

I can work on a PR if some support.

istarkov:
May be just check options on gestureHandling options, and depending on it just make off code section above. I cant help much here as have near zero knowledge about web on mobile. So feel free to provide a PR, will help if any questions I know the answer

zeeke:
I tried googling about maps and iPad issues but I find nothing.

@jooj123 can you give a feedback about the documentation i linked above?

jooj123:
@zeeke 

The ipad fix I added here: https://github.com/istarkov/google-map-react/pull/233
Prevents the whole page from being dragged on an ipad.
Without it the user experience is horrible on ipad.

Instead of doing this with `touchmove` events in react - it could possibly be done with "Scrolling and Panning" settings  (`greedy` option) with google maps.

But this will need to be tested - I have not checked if this prevents whats mentioned in https://github.com/istarkov/google-map-react/pull/233

yoadsn:
@zeeke I can see that you have forked this repo and reverted the PR for #233.
I tried the same locally and that indeed solves the problem.
@jooj123 @istarkov I think this change should be put behind a flag to be an opt-in. The reason is that google maps has it's own (complex) way of handling the duality of swipe over a map (scroll page? pan map?) with the `gestureHandling` option.

The only reason I'm not 100% sure this feature can be removed completely is due to handling of the rendered markers/layers on top of the gmaps. Perhaps this has no way of working with the `gestureHandling` feature - but currently it's really destroying this lib for any touch enabled use.

istarkov:
Lets remove, @yoadsn can you check that greedy option? Ill accept a PR with removal :-) but dont like the flag idea

yoadsn:
[...]

PR #334, opened 2017-03-10
Stop capturing all touch events and use built in google maps support

yoadsn:
This should solve #297 and reverts the changes from #233.
The purpose of this PR is to expose the built in google maps functionality for handling touch interaction with the map and removing the previous handling which captured all touch events and emulated a "greedy" approach where the other elements on the page cannot respond to those touch events.

### DIFF
--- a/API.md
+++ b/API.md
@@ -318,3 +318,21 @@ function createMapOptions() {
   };
 }
 ```
+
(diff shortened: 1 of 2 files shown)


istarkov:
Super, thank you!

PR #335, opened 2017-03-10
Fix unresponsive map after outside-click on a touch device with draggable:false

### LINKED ISSUE #332: Map not responsive after click outside the map when on touch device and draggable:false option
I may have found the wrong reproduction of this problem but I modified an old JSBin to show the problem:
Currently checked on chrome dev tools using the "mobile" tap input emulation.
https://jsbin.com/wihaxezera/edit?js,console,output

After the map loads - tapping the marker would produce a double "click" console log.. (should not be double)
Then, clicking on the red area to the right (just outside the map div container) would make click event to never fire again.

When removing either `draggable:false` or `fullscreenControl:false` this behaviour is not observed.

Did not have a chance to test yet on an actual mobile device..

yoadsn:
This fixes #332 with a workaround for the detection of the presence of the mouse in the map.
See the original issue for reproduction and root cause.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -645,10 +645,10 @@ export default class GoogleMap extends Component {
       });
 
       // an alternative way to know the mouse is back within the map
-      // This would not fire when clicking/interacting with google maps own on-map countrols+markers.
-      // This handles an edge case for touch devices + 'draggable:false' custom option.
-      // See #332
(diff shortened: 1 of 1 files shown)


istarkov:
Thank you!!! Will release on npm today.

PR #346, opened 2017-03-20
Support className prop

glennreyes:
Could be useful 😉

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -51,7 +51,7 @@ export default class GoogleMap extends Component {
   static propTypes = {
     apiKey: PropTypes.string,
     bootstrapURLKeys: PropTypes.any,
-
+    className: PropTypes.string,
     defaultCenter: React.PropTypes.oneOfType([
       PropTypes.array,
       PropTypes.shape({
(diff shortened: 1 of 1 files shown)


istarkov:
> Could be useful
Just one case I see, setting `flex:1` ;-) 

I think to remove even style property, also remove internal autoSizer, and to include AutoSizer from react-virtualized library which works well for block and flex containers

glennreyes:
What do you mean by `flex: 1`? 😄 
Imho the wrapper should be as intuitive as possible 👍  What I've just added here was the possibility to define CSS rules via a className. So it saves me one extra div

istarkov:
For me all that `style` and `className` exposing as a props, looks like an antipattern. The best what can do non styled component as like as map is to fit it's parent size. 

The problem that inside `display: flex` container defaults like `width: 100%; height: 100%` will not working, and even you set `flex:1` with `width` and `height` defined this won't help.

So in future version of map component I want to remove `style` prop, and provide an `AutoSizer` as a root map container.

And what the reason to remove additional `div` here, I could understand that for controls like `table` every additional cell `div` can cause performance issues but for controls like map even 10 additional divs around will not affect perf.

dverbovyi:
Hi, I need this feature as well. Is it going to be merged?

PR #400, opened 2017-06-20
Allow user to specify the position

SyedSaifAli:
It is occupying the whole area.
I just wanted to use the map within a region. Not the whole div.

### DIFF
--- a/src/google_map_map.js
+++ b/src/google_map_map.js
@@ -7,7 +7,7 @@ const style = {
   top: 0,
   margin: 0,
   padding: 0,
-  position: 'absolute',
+//   position: 'absolute',
 };
 
 export default class GoogleMapMap extends Component {


istarkov:
Within what? What the problem this change could solve?

SyedSaifAli:
Actually, currently the map is visible in the whole page.  Is it possible that I can use the style as I want? Like I don't want the position as absoulte because it is consuming the whole page. What other possible solutions are there?

istarkov:
The other option is to read css spec about position property.

SyedSaifAli:
And how can I do that?

SyedSaifAli:
Hey Ivan.. Your google-map-react module was very good. Can you please tell
me that by using that package can I draw some polylines as I get co
ordinates from the server in real time?

Thanks man

PR #440, opened 2017-09-04
Do a proper check for object equality

rmacqueen:
Previously we were just using a simple '==' to check if
new defaultCenter was equal to old defaultCenter. However,
defaultCenter is an object, and therefore '==' will always
return false unless they reference the same object in memory.
Instead we should do a proper comparison of the properties
of the object.

### DIFF
--- a/src/utils/objectEquality.js
+++ b/src/utils/objectEquality.js
@@ -1,0 +1,25 @@
+export default function areObjectsEqual(a, b) {
+    // Create arrays of property names
+    var aProps = Object.getOwnPropertyNames(a);
+    var bProps = Object.getOwnPropertyNames(b);
+
+    // If number of properties is different,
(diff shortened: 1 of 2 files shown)


src/utils/objectEquality.js
```diff
@@ -0,0 +1,25 @@
+export default function areObjectsEqual(a, b) {
+    // Create arrays of property names
+    var aProps = Object.getOwnPropertyNames(a);
```
itsmichaeldiego: Lets use const or let instead of var.

itsmichaeldiego:
@rmacqueen As you did not answer I handled this using facebook's `shallowEquals` method.

rmacqueen:
@itsmichaeldiego oops, sorry didn't see this. What you did works though. Thanks!

Issue #296, opened 2016-12-29
Markers position in fullscreen mode


nitroin:
Enable fullscreen control and control position with `options` prop:

```
const options = {
    fullscreenControl: true,
    fullscreenControlOptions: {
        position: 10
    }    
};

<GoogleMap
    center={this.props.center}
    zoom={this.props.zoom}
    options={options}
    >
    {this.renderMarkers()}
</GoogleMap>
```

**Expected:**
- [x] A React component should render with proper control for fullscreen mode, with markers in the correct position.
- [x] A click on the fullscreen control should render the component in fullscreen.
- [ ] Markers are correctly positioned while in fullscreen mode.

freecates:
Hi @nitroin , is this solved?

viperfx07:
> Markers are correctly positioned while in fullscreen mode.

still waiting for this to be fixed

istarkov:
@viperfx07 having that I'm not using this feature in any of my projects, I'm not interested to somehow fix it. Its opensource project, you are welcome to fix.

thg303:
I'm using version 0.31.0, problem still exists. markers do not move in fullscreen mode.

itsmichaeldiego:
@thg303 Please upload a snippet reproducing the issue, thanks

PR #452, opened 2017-10-06
Fix markers position on fullscreen mode

jbcochery:
On fullscreen mode, the view size should not be computed from the DOM element size, but from the `window` dimensions.

Fixes https://github.com/istarkov/google-map-react/issues/296

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -848,9 +848,16 @@ export default class GoogleMap extends Component {
 
   _setViewSize = () => {
     if (!this.mounted_) return;
-
-    const mapDom = ReactDOM.findDOMNode(this.googleMapDom_);
-    this.geoService_.setViewSize(mapDom.clientWidth, mapDom.clientHeight);
+    if (
+      document.fullscreen ||
+      document.webkitIsFullScreen ||
+      document.mozFullScreen
+    ) {
+      this.geoService_.setViewSize(window.innerWidth, window.innerHeight);
+    } else {
+      const mapDom = ReactDOM.findDOMNode(this.googleMapDom_);
+      this.geoService_.setViewSize(mapDom.clientWidth, mapDom.clientHeight);
+    }
     this._onBoundsChanged();
   };
 


src/google_map.js
```diff
@@ -848,9 +848,12 @@ export default class GoogleMap extends Component {
 
   _setViewSize = () => {
     if (!this.mounted_) return;
-
-    const mapDom = ReactDOM.findDOMNode(this.googleMapDom_);
-    this.geoService_.setViewSize(mapDom.clientWidth, mapDom.clientHeight);
+    if (document.fullscreen || document.webkitIsFullScreen || document.mozFullScreen) {
+      _this.geoService_.setViewSize(window.innerWidth, window.innerHeight);
```
istarkov: _this is not defined

emilpalsson:
Great find @jbcochery! I was just about to start looking for a fix for #296 when I saw your PR. Crazy timing, for an issue 10 month old. Thanks @istarkov for a great lib!

src/google_map.js
jbcochery: Fixed, sorry about that!

itsmichaeldiego:
Wow @emilpalsson same here! I was looking at that issue right now.

This is looking great @jbcochery, thank you! And thank you @istarkov for the lib!

nicolasdelfino:
Any status of when this is getting merged?

itsmichaeldiego:
@istarkov How are we looking on merging this? 👍

qur2:
I also got bitten by that issue. While the fix offered seems to work (thanks for that!), the real problem is that the ref that GooleMapReact keeps does not point to the div used to render the fullscreen map. The fix works because incidentally, the div to which we should have a ref to has the same size than the window.

Venumteam:
@istarkov Hey, how about merging this fix?

ShepelievD:
@istarkov really need this stuff, merge, please! 😢 😢 😢

nicolasdelfino:
@istarkov Please communicate if this is getting merged or not. It would be a shame if people started working on their own forks and this project would stagnate.

jkettmann:
@istarkov We also need this fix. Do you plan on merging this?

istarkov:
Guys I need collaborator-leader of this project, see the readme top. I have no desire now to do anything :-(

nicolasdelfino:
@istarkov I understand your position and hope you'll find someone willing to help. I'm sure you'll do. 
In the meantime there is no reason for you not to merge this.

src/google_map.js
```diff
@@ -848,9 +848,16 @@ export default class GoogleMap extends Component {
 
   _setViewSize = () => {
     if (!this.mounted_) return;
-
-    const mapDom = ReactDOM.findDOMNode(this.googleMapDom_);
-    this.geoService_.setViewSize(mapDom.clientWidth, mapDom.clientHeight);
+    if (
+      document.fullscreen ||
+      document.webkitIsFullScreen ||
+      document.mozFullScreen
```
itsmichaeldiego: Could we set up a function called `isFullScreen` and reuse it?

itsmichaeldiego: Could we create a function called `isFullScreen` instead?

itsmichaeldiego:
This is merged and published. Feel free to upgrade to `0.26.0`

CC @nicolasdelfino @emilpalsson @minhkl @huwr @paupalou @jkettmann @ShepelievD @Venumteam @qur2 @jbcochery

itsmichaeldiego:
@jbcochery There is a little issue with this PR, could you take a look? https://github.com/istarkov/google-map-react/issues/539#issuecomment-375767607

PR #461, opened 2017-11-07
Remove throw e in google_map.js

mistaguy:
remove `throw e` to allow other reach component to better handle the error in react 15. Having this error makes it difficult for other component to handle this error and probably show their own UI since the execution stops at the line that throws the error


mistaguy:
I have a work around. I will post it in another pull request

PR #472, opened 2017-11-26
Fixed typo

etienne-martin:


itsmichaeldiego:
@etienne-martin Thanks for your PR, Readme has been updated and this is no longer needed.

PR #479, opened 2017-12-20
notify call back of load failure

mistaguy:
When the google  fails to load due to no internet, the  call back should be notified so that appropriate UI can be shown


mistaguy:
@itsmichaeldiego thanks for spotting the bug 🐛  . It is because i edited the file on github and not my work station but am glad that the unit tests got me :) i have done a forced push with the correction

PR #482, opened 2017-12-21
Make `resetBoundsOnResize` preserve center when full-screened

josephfrazier:
This supersedes and resolves https://github.com/istarkov/google-map-react/pull/308,
as specified in https://github.com/istarkov/google-map-react/pull/308#discussion_r158299572


src/google_map.js
```diff
@@ -398,7 +399,7 @@ export default class GoogleMap extends Component {
     window.removeEventListener('keydown', this._onKeyDownCapture);
     mapDom.removeEventListener('mousedown', this._onMapMouseDownNative, true);
     window.removeEventListener('mouseup', this._onChildMouseUp, false);
-    if (this.props.resetBoundsOnResize) {
+    if (this.props.resetBoundsOnResize || this.props.lockCenter) {
```
itsmichaeldiego: Wondering if `|| this.props.lockCenter` is really needed, as `resetBoundsOnResize` should be set to true if we want to lock the center, not sure if I am explaining myself, but I believe the map always should re-set the center when reseting the bounds on resize.

josephfrazier: Hmm, I think I see what you're saying. I hadn't used `resetBoundsOnResize` before, but I just tried it out on the master branch and on this PR's branch by running `yarn start` and visiting http://localhost:4000/resizable. On both branches, resizing by dragging the lower- the map keeps the map centered, but only with this PR does clicking the fullscreen button also keep the map centered. I think it makes sense for `resetBoundsOnResize` to include the `lockCenter` behavior, so I removed the `lockCenter` stuff while keeping the behavioral changes. See https://github.com/istarkov/google-map-react/pull/482/commits/692e15b9b19dcb65d85614e7f6e54f6f7ba04188

itsmichaeldiego:
@josephfrazier Just to confirm, did you test latest changes?

josephfrazier:
Yes, I verified the fix as outlined [above](https://github.com/istarkov/google-map-react/pull/482#discussion_r158341280), by running `yarn start`, then visiting http://localhost:4000/resizable and clicking the full-screen button.

PR #483, opened 2017-12-21
Update yarn.lock

josephfrazier:
https://github.com/istarkov/google-map-react/commit/e344c3c8f4e51f6edea0d0f2d437f2bee661220d switched from `point-geometry` to `@mapbox/point-geometry`, but it didn't update yarn.lock accordingly. This does.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,10 @@
 # yarn lockfile v1
 
 
+"@mapbox/point-geometry@^0.1.0":
+  version "0.1.0"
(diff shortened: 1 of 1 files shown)


josephfrazier:
Thanks! If you'd like, I can add a command to `.travis.yml` that ensures `yarn.lock` stays updated. See here for an example: https://github.com/OctoLinker/browser-extension/blob/f2b242cdfbacaeeb6c973ae791fb01b8aabba7de/.travis.yml#L20-L21

itsmichaeldiego:
@josephfrazier That would be great! Feel free to do so and I'll approve.

PR #494, opened 2018-01-16
Fix for usage from China

### LINKED ISSUE #493: Usage from China requires different API URL
If using Google Maps APIs from China a different URL should be used, see https://developers.google.com/maps/faq#china_ws_access.

https://maps.googleapis.com should be replaced with http://maps.google.cn, since the regular one is not accessible from within China.

emilpalsson:
Use China specific API URL if setting bootstrapURLKeys.region to 'cn'.

Fixes #493

### DIFF
--- a/src/utils/loaders/google_map_loader.js
+++ b/src/utils/loaders/google_map_loader.js
@@ -62,8 +62,10 @@ export default function googleMapLoader(bootstrapURLKeys) {
       ''
     );
 
+    var url = bootstrapURLKeys.region === 'cn' ? 'http://maps.google.cn' : 'https://maps.googleapis.com'
+
     $script_(
-      `https://maps.googleapis.com/maps/api/js?callback=_$_google_map_initialize_$_${queryString}`,
+      `${url}/maps/api/js?callback=_$_google_map_initialize_$_${queryString}`,
       () =>
         typeof window.google === 'undefined' &&
         reject(new Error('google map initialization error (not loaded)'))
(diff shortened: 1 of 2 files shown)


itsmichaeldiego:
@emilpalsson Thanks for this, it is looking good, would you mind adding this to the docs? As soon as you do that I will release a new version fixing this.

src/utils/loaders/google_map_loader.js
```diff
@@ -62,8 +62,12 @@ export default function googleMapLoader(bootstrapURLKeys) {
       ''
     );
 
+    const url = bootstrapURLKeys.region === 'cn'
```
jooj123: what about other regions ?

itsmichaeldiego: @jooj123 I believe the only region that has a problem with this is China. But we could confirm

emilpalsson: Exactly, China is the only region mentioned on the Google Maps FAQ (see link in the issue), and the only region I've experienced problems with.

emilpalsson:
@itsmichaeldiego Great! Of course, I'll see if I can find a good spot for this in the docs right now.

emilpalsson:
@itsmichaeldiego I've added this to the docs, let me know if you want me to structure it differently. I also added a toLocaleLowerCase() for case insensitivity.

romanenko:
Looks like that if region is not set — it leads to uncaught Error now, with 0.28.0 release

image

I'm setting up my GoogleMap component like this:

```js
      <GoogleMap
        bootstrapURLKeys={{
          key: GOOGLE_MAPS_API_KEY,
          language: intl.locale,
        }}
//...
```

emilpalsson:
@romanenko Yes, I'm sorry for this. There is already a new PR to fix this (#494).

PR #496, opened 2018-01-19
Fix China fix making region required

emilpalsson:
Bah, sorry for stupid misstake. My PR (#494) to fix (#493) made the code break if not specifying bootstrapURLKeys.region. Sorry, here is a fix for that.

### DIFF
--- a/src/utils/loaders/google_map_loader.js
+++ b/src/utils/loaders/google_map_loader.js
@@ -62,7 +62,8 @@ export default function googleMapLoader(bootstrapURLKeys) {
       ''
     );
 
-    const url = bootstrapURLKeys.region.toLocaleLowerCase() === 'cn'
+    const url = bootstrapURLKeys.region &&
+      bootstrapURLKeys.region.toLowerCase() === 'cn'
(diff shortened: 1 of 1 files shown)


pkeuter:
Thanks for this! Please merge ASAP because the current release breaks on the region being undefined.

emilpalsson:
@itsmichaeldiego Please merge this, as the current release is not working.

aldoso:
can someone please explain how is this working or if there is any proper documentation. It seems it's not woking doing what it says on the readme :(

emilpalsson:
@aldoso What problems do you experience? The only thing this PR (together with #494) does is making the lib use the China API URL if setting bootstrapURLKeys.region to 'cn'. If not setting 'cn' there should be no difference to version 0.27.0 (the version prior to these two PRs).

This is explained [here](https://github.com/istarkov/google-map-react/blob/master/API.md#localizing-the-map) in the docs.

However, a bug made version 0.28.0 break if not specifying a region at all. This is no longer a problem in 0.29.0.

Issue #476, opened 2017-12-14
IE10 Unable to get property 'center' of undefined or null reference


falconmick:
This issue occurred before in #223 and is back with a vengeance... 
On only IE10 this is happening:
IE10 Unable to get property 'center' of undefined or null reference

code that causes issue:

	    _this.mounted_ = false;
	    _this.initialized_ = false;
	    _this.googleApiLoadedCalled_ = false;
	
	    _this.map_ = null;
	    _this.maps_ = null;
	    _this.prevBounds_ = null;
	
	    _this.layers_ = {};
	
	    _this.mouse_ = null;
	    _this.mouseMoveTime_ = 0;
	    _this.boundingRect_ = null;
	    _this.mouseInMap_ = true;
	
	    _this.dragTime_ = 0;
	    _this.fireMouseEventOnIdle_ = false;
	    _this.updateCounter_ = 0;
	
	    _this.markersDispatcher_ = new _marker_dispatcher2.default(_this);
	    _this.geoService_ = new _geo2.default(K_GOOGLE_TILE_SIZE);
            // this line under me
	    _this.centerIsObject_ = (0, _is_plain_object2.default)(_this.props.center);
	
	    _this.minZoom_ = DEFAULT_MIN_ZOOM;
	    _this.defaultDraggableOption_ = true;
	
	    _this.zoomControlClickTime_ = 0;
	
	    _this.childMouseDownArgs_ = null;
	    _this.childMouseUpTime_ = 0;
	
	    _this.googleMapDom_ = null;

I don't really know how to add map file support for IE10 so I have no idea where this is :( any ideas?

falconmick:
I have tried supplying center prop to google maps but still no use

falconmick:
@istarkov you still active?

itsmichaeldiego:
@falconmick Sorry for the delay, did #504 being merged fix your problem?

falconmick:
Sorry for not replying till now, yeah sure did

itsmichaeldiego:
Great @falconmick, I am closing the issue then.

Thanks for your contribution ♥️

PR #504, opened 2018-01-29
Added loose to es2015 for IE10, IE11 support

falconmick:
IE9 and IE10 require loose to work
    
your demo app doesn't work on IE9/10 so you will need to make a more simple one to test this fix out. the issue is to do with this.props not being avaiable in constructors for IE9/10 see: https://github.com/babel/babel/issues/3975 it was supposidly fixed 1 year ago but I can't see it working, maybe we need to swap to @babel/core to get this fix? I tried it but the change is made larger due to the curent demo apps build system.
    
Full fix would involve upgading to @babel/core + @babel/preset-env and migrating away from kotatsu in favour of a plain ole react app via webpack (create-react-app maybe?) and then place it into another folder and have that folder habing it's own package.json so we can consume google-map-react as if it were a real library during testing. Sure we would loose hmr, but at this stage hmr isn't all that useful anyway.
    
I'de love the proper fix, but for now I have clients with IE10 browser requirements.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,3 @@
 {
-  "presets": ["es2015", "react", "stage-0"]
+  "presets": [["es2015", {"loose": true}], "react", "stage-0"]
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 lib/*
 dist/*
 /node_modules
+.idea/


falconmick:
fixes #476

falconmick:
this is the PR in which babel was supposed to fix this, I'm not sure how to verify if babel-core will get these updates if we upgrade it (I ran yarn upgrade but I didn't update the version number so I think the carrot limits us to A.x.x) but it looks like it's a @babel/core fix
https://github.com/brandleadership/partystreusel/pull/51

itsmichaeldiego:
@falconmick That is a great input, thanks for that. To be honest I am not sure if we would like to invest that much time taking care about iE 9 and iE10, as they're close to be outdated (hopefully). This ofc if we're talking just about the examples, I believe it is important to be able to support at least iE10, so this PR is 👍

Issue #490, opened 2018-01-05
Incorrect marker animations with google maps API beta renderer


stephenfarrar:
There is a beta renderer for the google maps API
https://developers.google.com/maps/documentation/javascript/beta-renderer

With the beta renderer, the react markers behave very strangely when zooming in. Best I can tell, this is for two reasons:

1. The new renderer uses fractional zoom levels during the zoom animation. But `map.getZoom()` doesn't return the fractional zoom. Instead, it's reflected in the projection, and factored into methods like `.fromLatLngToDivPixel()`.

2. The new renderer calls `OverlayView`'s `.draw()` method on every frame of the animation. But it seems like this library draws its markers at some other point in time — not during `.draw()`.

For an example of how `OverlayView` is properly used to draw markers, see https://developers.google.com/maps/documentation/javascript/examples/overlay-popup

jabidof:
Hello,
I think this really brings strange user experience.
Is there a way to use previous version of GMap API?

itsmichaeldiego:
@jabidof @stephenfarrar Solution has been provided here: https://github.com/istarkov/google-map-react/issues/510#issuecomment-366002128

PR #530, opened 2018-03-10
Set latest release version as default if user does not specify a version.

### LINKED ISSUE #510: Zooming is animating markers from the corners
I checked the [demo app](http://istarkov.github.io/google-map-react/map/main/) and it's behaving the same way when I change zoom. I suppose it's not a feature, right?

### LINKED ISSUE #526: Markers are not clickable on mobile
First i thought it may be something wrong with my code , but when i check demos with my mobile phone i saw it has the same problem , markers are not clickable on mobile phone !!? And every things works fine on desktop.

itsmichaeldiego:
Closes #510
Closes #526 

There are a couple of bugs involved in the latest release of Google Maps, both issues (#510 and #526) were caused by the version `3.32` which is **experimental**. [Quoting Google](https://developers.google.com/maps/documentation/javascript/versions):

> The experimental version — currently 3.32 — contains the latest features and bug fixes as they are made publicly available. Changes made to the experimental version are not guaranteed to be feature stable. We encourage you to regularly test your applications with the experimental version, which you can do by adding v=3.32 when loading the API

We would like our default version to be the latest **release** version, and not the latest **experimental** version. To do so we set up `v=3` if no version is specified by the user. Which will be, today 3/10/2018, `3.31`:

> The current release version is 3.31. You can request it with either of the following bootstraps:

```
<script async defer
    src="https://maps.googleapis.com/maps/api/js?v=3
        &key=YOUR_API_KEY&callback=initMap">
```
```
<script async defer
    src="https://maps.googleapis.com/maps/api/js?v=3.31
        &key=YOUR_API_KEY&callback=initMap">
```

I am going to use `3.31` to make sure that this library works, when `3.32` is released, we would need to check if the issues that are affecting us now, will still be affecting us then. But for now, I think this is a good solution, this is also recommended by Google:

> In order to make sure that there are no issues arising from version rollover, we recommend that you explicitly specify the number of the current release version of the API in the bootstrap. For example, v=3.31.

Source: https://developers.google.com/maps/documentation/javascript/versions

### DIFF
--- a/src/utils/loaders/google_map_loader.js
+++ b/src/utils/loaders/google_map_loader.js
@@ -1,15 +1,26 @@
 /* eslint-disable no-console */
+const BASE_URL = 'https://maps';
+const DEFAULT_URL = `${BASE_URL}.googleapis.com`;
+
+const getUrl = region => {
+  if (region && region.toLowerCase() === 'cn') {
+    return `${BASE_URL}.google.cn`;
+  }
+  return DEFAULT_URL;
+};
+
 let $script_ = null;
 
 let loadPromise_;
 
 let resolveCustomPromise_;
+
 const _customPromise = new Promise(resolve => {
   resolveCustomPromise_ = resolve;
 });
 
 // TODO add libraries language and other map options
-export default function googleMapLoader(bootstrapURLKeys, heatmapLibrary) {
+export default (bootstrapURLKeys, heatmapLibrary) => {
   if (!$script_) {
     $script_ = require('scriptjs'); // eslint-disable-line
   }
@@ -23,6 +34,7 @@ export default function googleMapLoader(bootstrapURLKeys, heatmapLibrary) {
   if (loadPromise_) {
     return loadPromise_;
   }
+
   loadPromise_ = new Promise((resolve, reject) => {
     if (typeof window === 'undefined') {
       reject(new Error('google map cannot be loaded outside browser env'));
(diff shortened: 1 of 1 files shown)


domesticmouse:
As a Googler on the Google Maps Developer Relations team, I can categorically state that this isn't a solution, just a delaying tactic. The advice given by our Google Maps JavaScript API team lead @stephenfarrar in https://github.com/istarkov/google-map-react/issues/490 is coming from a place of knowledge. He looks after the API, after all.

I strongly suggest that you investigate fixing this properly by following his advice, and building on the prototype he provided in https://github.com/istarkov/google-map-react/issues/510#issuecomment-370145748

stephenfarrar:
FYI #526 is now fixed in v=3.32

itsmichaeldiego:
@domesticmouse Thanks for your message, as a Collaborator on the Open Source Team, I try to unblock current or new users providing a solution from our side. This was totally temporary and I never mentioned that this was the final solution, as I stated:

> when 3.32 is released, we would need to check if the issues that are affecting us now, will still be affecting us then. But for now, I think this is a good solution, this is also [recommended by Google](https://developers.google.com/maps/documentation/javascript/versions#for-high-availability--high-traffic-applications)

Right now as per @stephenfarrar's [comment](https://github.com/istarkov/google-map-react/pull/530#issuecomment-372872802). This is fixed and, we do not need to change our code.  So, I believe this was a good call.

Thanks both of you @stephenfarrar and @domesticmouse

stephenfarrar:
@itsmichaeldiego Thanks for clarifying. Requesting 3.31 is exactly what we hope people will do if they encounter bugs in the Maps API like #526.

On the other hand, #490 and its duplicate #510 are not bugs in the Maps API. Rather, I think this library is doing something we don't support (all sorts of assumptions about how animations used to be implemented). I'm hoping that we could work together to remove these assumptions and make the library work even better.

I wasn't sure whether anyone was intending to keep on working at #510, given that the issue was closed. Should we open a new issue?

itsmichaeldiego:
@stephenfarrar Opening a new issue sounds good, I am also looking forward to fix this! Feel free to message me and we can pair up!

PR #533, opened 2018-03-11
Refactor code style

itsmichaeldiego:
This pr includes:

- Fix some eslint styling
- Update recompose to latest version
- Use object destruction to import multiple child modules from one single module (recompose & lodash)

PR #534, opened 2018-03-11
Use specific packages from lodash instead of whole library

itsmichaeldiego:
This PR is just one proposal to use specific packages from lodash instead of importing the whole library.

I came across with the problem that we tried to remove lodash here: https://github.com/istarkov/google-map-react/commit/670fadd2da1086bc41185a00503280929e4476a1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R72 but we actually moved to devDependencies.

But accidentally, because it was there, @ZAKdev used `fp` from lodash, as it was still in the devdependencies, in #441.

After that, I also used isEmpty from lodash in #533 

This increased the build size to 797kb standard and 201kb minified.

This PR decreases it to 398kb and 93.1kb minified. 

Output: Size decreased by more than 50%

## Prev:
screen shot 2018-03-11 at 2 00 52 pm

## Now:
screen shot 2018-03-11 at 1 59 57 pm


src/google_heatmap.js
```diff
@@ -1,8 +1,9 @@
-import fp from 'lodash/fp';
+import map from 'lodash.map';
+import reduce from 'lodash.reduce';
```
ZAKdev: Instead of importing functions map and reduce from lodash, we can simply call native map reduce functions. If you agree i can change in this PR

itsmichaeldiego: @ZAKdev If you can make a separate PR doing that, I can handle the rest. This is my first option, but the second one would be using all _our own utils_ instead. So for both options the change you're proposing would be great.

itsmichaeldiego: @ZAKdev Either way I go, your change independently will decrease the size of the build.

ZAKdev: @itsmichaeldiego, Ok i will do it in separate PR, i will open a PR soon and let me know if i can help you more to decrease the size of build.

itsmichaeldiego: @ZAKdev Thanks, it will be really helpful for you to change the reduce and map given that you can test the changes in the heatmap and see if it works, thanks for that! I will be waiting on this, do you think you can do it today? ♥️

itsmichaeldiego:
I am going with solution #535 as it creates a smaller build, if required in the future we could move to this approach.

cc @ZAKdev

itsmichaeldiego:
I am going for #535 as for now.

PR #535, opened 2018-03-12
Remove lodash and use our own utils

itsmichaeldiego:
This PR is just one proposal to use specific packages from lodash instead of importing the whole library.

I came across with the problem that we tried to remove lodash here: https://github.com/istarkov/google-map-react/commit/670fadd2da1086bc41185a00503280929e4476a1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R72 but we actually moved to devDependencies.

But accidentally, because it was there, @ZAKdev used `fp` from lodash, as it was still in the devdependencies, in #441.

After that, I also used isEmpty from lodash in #533 

This increased the build size to 797kb standard and 201kb minified.

This PR decreases it to 140kb and 47.5kb minified. 

Output: Size decreased by almost 4 times.

## Prev:
screen shot 2018-03-11 at 2 00 52 pm

## Now:
screen shot 2018-03-11 at 7 43 57 pm


itsmichaeldiego:
@ZAKdev Let me know if this change works. Otherwise feel free to open a PR or add a commit here, I can go either way.

src/google_heatmap.js
```diff
@@ -1,21 +1,15 @@
...
+      []
     ),
   });
 
 export const optionsHeatmap = (instance, { options }) =>
-  fp.map(
-    option => instance.set(option, options[option]),
-    Object.keys(options || {})
-  );
+  Object.keys(options).map(option => instance.set(option, options[option]));
```
ZAKdev: I don't know why but i am not able to run entire project, so everything looks good to me just map over an empty object as well if there is no options, like below.

```Object.keys(options || {}).map(option => instance.set(option, options[option]));```

ZAKdev:
This is what i am getting from every branch.

screen shot 2018-03-12 at 19 35 13

itsmichaeldiego:
@ZAKdev Fixed this in #537, you should be looking good now, please start from *latest master*

src/google_heatmap.js
itsmichaeldiego: @ZAKdev Done, instead of doing what you said I've set a default value for the parameter. LMK How that works. I am merging this as for now.

ZAKdev:
@itsmichaeldiego, i can't push any of my code, seems like not having access.

itsmichaeldiego:
@ZAKdev sorry, push to where?

ZAKdev:
@itsmichaeldiego, i added weight in my heapmap implementation but i am not able to push my code to this repo.

Do we have any slack channel where we can comminicate or via email, here is my email zainahmedkhan@gmail.com

PR #559, opened 2018-04-13
Fix zoom animation for v3.32

### LINKED ISSUE #552: Fix zoom animation with Google Maps API v3.32
This is a new issue to discuss a long-term solution to the zoom animation problem that started with v3.32 of the Google Maps API. See also #490 #510 #530 

I've begun work in https://github.com/stephenfarrar/google-map-react/tree/zoom-animation
Someone could start a branch here, and I'll create a pull request.

So far, here are some of the issues I'm confused about:

* The markers are children of `overlay.div` in `src/google_map.js` . That div used to be positioned & sized so that it was the size of the map. I don't think that's necessary, and I've removed it, but perhaps it's causing problems elsewhere.

* I'm not totally sure what coordinate system is used in the geo service `project()` and `unproject()` when `projectFromLeftTop` is false. Did it mean 'relative to the map center'? I've made it mean 'relative to the MapPanes'. This makes sense for markers, but I'm not sure that it makes sense elsewhere.

* My branch has problems with markers disappearing and reappearing. Perhaps this indicates some problems with keeping track of the current map bounds.

stephenfarrar:
I managed to reduce the surface of the change.
The basic demo seems to work fine, but I haven't tried it in any other demos.

Summary of change:

* This uses a better method to position the marker container div in the map top-left.
* It also lets each marker use the `fromLatLngToContainerPixel()` method, which accounts for fractional zoom during a zoom animation.

Fixes #552

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -34,6 +34,9 @@ const K_GOOGLE_TILE_SIZE = 256;
 const K_IDLE_TIMEOUT = 100;
 const K_IDLE_CLICK_TIMEOUT = 300;
 const DEFAULT_MIN_ZOOM = 3;
+// Starting with version 3.32, the maps API calls `draw()` each frame during
+// a zoom animation.
+const DRAW_CALLED_DURING_ANIMATION = 32;
 
 function defaultOptions_(/* maps */) {
   return {
@@ -566,8 +569,10 @@ export default class GoogleMap extends Component {
 
         this._setLayers(this.props.layerTypes);
 
+        // Parse `google.maps.version` to capture the major version number.
         const versionMatch = maps.version.match(/^3\.(\d+)\./);
-        const mapsVersion = versionMatch ? Number(versionMatch[1]) : 99;
+        // The major version is the first (and only) captured group.
+        const mapsVersion = versionMatch && Number(versionMatch[1]);
 
         // render in overlay
         const this_ = this;
@@ -663,7 +668,9 @@ export default class GoogleMap extends Component {
               this_._onZoomAnimationStart();
             }
 
-            if (mapsVersion < 32) {
+            // If draw() is not called each frame during a zoom animation,
+            // simulate it.
+            if (mapsVersion < DRAW_CALLED_DURING_ANIMATION) {
               const TIMEOUT_ZOOM = 300;
 
               if (
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -591,9 +591,10 @@ export default class GoogleMap extends Component {
 
             const panes = this.getPanes();
             panes.overlayMouseTarget.appendChild(div);
-            if (mapsVersion >= 32) {
-              this_.geoService_.setProjection(maps, overlay.getProjection());
-            }
+            this_.geoService_.setMapCanvasProjection(
+              maps,
+              overlay.getProjection()
+            );
 
             ReactDOM.unstable_renderSubtreeIntoContainer(
               this_,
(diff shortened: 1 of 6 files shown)


src/loaders/google_map_loader.js
```diff
@@ -72,11 +72,9 @@ export default (bootstrapURLKeys, heatmapLibrary) => {
       ''
     );
 
-    // if no version is defined, we want to get the release version
-    // and not the experimental version, to do so, we set v=3.31
-    // src: https://developers.google.com/maps/documentation/javascript/versions
+    // See https://developers.google.com/maps/documentation/javascript/versions
     if (isEmpty(bootstrapURLKeys.v)) {
-      params += '&v=3.31';
+      params += '&v=3.exp';
```
itsmichaeldiego: 👍

itsmichaeldiego: 👍

src/google_map.js
```diff
@@ -659,25 +663,27 @@ export default class GoogleMap extends Component {
...
-              // but reduce it's occurence probability
-              raf(() =>
-                raf(() => {
-                  this_.updateCounter_++;
-                  this_._onBoundsChanged(map, maps);
-                }));
-            } else {
-              this_.updateCounter_++;
-              this_._onBoundsChanged(map, maps);
+            if (mapsVersion < 32) {
```
itsmichaeldiego: Can we move 32 to a const variable?

src/google_map.js
```diff
@@ -718,11 +724,8 @@ export default class GoogleMap extends Component {
           const div = overlay.div;
           const overlayProjection = overlay.getProjection();
           if (div && overlayProjection) {
-            const bounds = map.getBounds();
-            const ne = bounds.getNorthEast();
-            const sw = bounds.getSouthWest();
             const ptx = overlayProjection.fromLatLngToDivPixel(
-              new maps.LatLng(ne.lat(), sw.lng())
+              overlayProjection.fromContainerPixelToLatLng({ x: 0, y: 0 })
```
itsmichaeldiego: Could you elaborate why setting `{ x:0, y:0 }` fixes the animation problem?

src/google_map.js
```diff
@@ -566,6 +566,9 @@ export default class GoogleMap extends Component {
 
         this._setLayers(this.props.layerTypes);
 
+        const versionMatch = maps.version.match(/^3\.(\d+)\./);
+        const mapsVersion = versionMatch ? Number(versionMatch[1]) : 99;
```
itsmichaeldiego: Could we make this code more readable? What is `99`? or `versionMatch[1]`? (I understand ofc but its not so readable)

stephenfarrar: Done.

src/google_map.js
stephenfarrar: Good question, this is subtle. There are two ways of getting the lat-lng coordinates of the map's top-left corner.

1. Using `map.getBounds()` gives you the *final bounds*. That's because it is based on `map.getCenter()` and `map.getZoom()` which give you the final center and the final zoom. In other words, if you call `map.setZoom()` or click on the zoom control, `map.getZoom()` immediately reflects the new zoom, and so does `map.getBounds()`. You would use this if you were, say, fetching data from a server based on the map bounds.

2. Using `overlayProjection.fromContainerPixelToLatLng({ x: 0, y: 0})` gives you the bounds for *the current frame*. We use this because we are drawing each frame.

src/google_map.js
```diff
@@ -34,6 +34,9 @@ const K_GOOGLE_TILE_SIZE = 256;
 const K_IDLE_TIMEOUT = 100;
 const K_IDLE_CLICK_TIMEOUT = 300;
 const DEFAULT_MIN_ZOOM = 3;
+// Starting with version 3.32, the maps API calls `draw()` each frame during
+// a zoom animation.
+const DRAW_CALLED_DURING_ANIMATION = 32;
```
itsmichaeldiego: @stephenfarrar Good work, I will add `_VERSION` to this const real quick if you don't mind.

src/google_map.js
itsmichaeldiego: @stephenfarrar Great, thanks for the answer. That makes sense, if `map.getBounds` gives you the _final bounds_, that is exactly what is making the markers animate from the corners.

itsmichaeldiego:
@stephenfarrar Although I keep thinking if we want to keep simulating the `draw()` for versions below `3.32`, as this map uses the latest version as default 🤔. We could remove this whole block: https://github.com/google-map-react/google-map-react/pull/559/files#diff-0146a78a4269b3d2562b8f44d0137200R673

What do you think?

stephenfarrar:
I wrote it this way because I think it's good for customers to be able to choose independently the version of the Maps API and google-map-react.

But I agree that this version-detection is not very good.

We can delete this section in late August 2018, because v3.31 is scheduled to be deleted by then. Does that sound acceptable?

itsmichaeldiego:
@stephenfarrar I think it does! Lets keep it as it is right now, I will delete it afterwards if needed.

PR #562, opened 2018-04-18
weights for heatmap

nilutz:
I added the weights to the google_heatmap.js file now individiual weights can be loaded onto the heatmap...


itsmichaeldiego:
This was handled in #572

itsmichaeldiego:
@nilutz We took care of it on v1.0.1 and its published, feel free to upgrade and let me know if you find any issues.

PR #570, opened 2018-05-03
Avoid null error

asundlihardig:
Avoid the error "Cannot read property 'removeEventListener' of null"

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -396,9 +396,11 @@ export default class GoogleMap extends Component {
   componentWillUnmount() {
     this.mounted_ = false;
     const mapDom = ReactDOM.findDOMNode(this.googleMapDom_);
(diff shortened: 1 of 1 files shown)


itsmichaeldiego:
@asundlihardig Could you provide a screenshot showing the error ?

asundlihardig:
I'm sorry @itsmichaeldiego . The error was reported through Sentry.io

itsmichaeldiego:
@asundlihardig Sorry I can't merge this without you showing me how this is blocking you or where is the error. There are thousands of places in the code where we could add defensive programming, I need to know why adding it here would make the difference.

itsmichaeldiego:
@asundlihardig Looking forward to hear back from you! Would like to merge this if its an issue for the users.

PR #580, opened 2018-05-14
Option to position and size components with two corners.

mtogstad:
This lets you lock a component to a specific bounding rectangle, which allows for precise tiling during zoom with the experimental google maps API.

I work on a project that uses a technique similar to the thousands-of-markers sample project linked to from the readme.  If you run that project against the 3.32 google maps API you can see the same issue I was having.  The tiles are not scaled when frames are drawn during a zoom, which causes the edges to either have gaps between them, or to overlap.  This adds the option to specify a second LatLng (the southeast corner) to give a component a bounds-locked box to render into.

Here is a before example:
giphy

And after:
giphy-1

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -274,6 +274,9 @@ export default class GoogleMapMarkers extends Component {
           top: pt.y,
         };
 
+        // If the component has a southeast corner defined (either as a LatLng, or a separate
+        // lat and lng pair), set the width and height based on the distance between the northwest
+        // and the southeast corner to lock the overlay to the correct geographic bounds.
         if (
           child.props.seLatLng !== undefined ||
           (
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -269,10 +269,29 @@ export default class GoogleMapMarkers extends Component {
           ? this.props.geoService.fromLatLngToContainerPixel(latLng)
           : this.props.geoService.project(latLng);
 
-        const stylePtPos = {
+        let stylePtPos = {
           left: pt.x,
           top: pt.y,
         };
+
+        if (
+          child.props.seLatLng !== undefined ||
+          (
+            child.props.seLat !== undefined &&
+            child.props.seLng !== undefined
+          )
+        ) {
+          const seLatLng = child.props.seLatLng !== undefined
+            ? child.props.seLatLng
+            : {lat: child.props.seLat, lng: child.props.seLng};
+
+          const sePt = this.props.projectFromLeftTop
+            ? this.props.geoService.fromLatLngToContainerPixel(seLatLng)
+            : this.props.geoService.project(seLatLng);
+
+          stylePtPos.width = sePt.x - pt.x;
+          stylePtPos.height = sePt.y - pt.y;
+        }
 
         let dx = 0;
         let dy = 0;


src/google_map_markers.js
```diff
@@ -269,11 +269,30 @@ export default class GoogleMapMarkers extends Component {
...
-        const stylePtPos = {
+        let stylePtPos = {
           left: pt.x,
           top: pt.y,
         };
 
+        if (
+          child.props.seLatLng !== undefined ||
+          (
+            child.props.seLat !== undefined &&
```
itsmichaeldiego: @mtogstad Could you explain briefly what are you doing here and why? Maybe leave a comment

itsmichaeldiego:
@stephenfarrar What do you think about this? Given its part of the code you've been working with.

stephenfarrar:
Looks good to me.

mtogstad:
Just pushed an explanation, and thanks for looking at this pull request so fast!

itsmichaeldiego:
@mtogstad Thank you very much for this! Please update to `1.0.3` and let me know how it works. Feel free to make more PRS, we need more help!

palashkaria:
@mtogstad Hey, I know that this issue is closed, but I'm curious, how did this help with the canvas markers? I am using it for a similar use case & facing the exact same problem. Could you share the relevant change you made in this repo, which fixed the issue? https://github.com/istarkov/google-map-thousands-markers

palashkaria:
I'm on 1.0.5 itself, but the issue is not with this lib, it's with the canvas implementation (https://github.com/istarkov/google-map-thousands-markers). 

What I understood from this change is that I have to pass `seLatLng` to `CanvasTile` in that implementation. But that doesn't seem to work..

itsmichaeldiego:
@palashkaria Oh yes, what changed is that you can pass either `seLatLng={[lat, lng]}` or `seLat={lat} seLng={lng}`

palashkaria:
I think it's like `seLatLng={{lat, lng}}`. But question is, specifically how does it help in those before/after images posted above, from the 'google-map-thousands-markers' example. I'm trying something here, but it doesn't seem to work..

https://github.com/palashkaria/google-map-thousands-markers/tree/google-maps-v3.32-fix

itsmichaeldiego:
@palashkaria What are you trying? Could you provide a live example?

itsmichaeldiego:
@palashkaria That should work given @mtogstad's change, but there might be a chance that you dont need that `seLatLng` anymore as we went through some interesting changes in the way markers are drawn. What exactly you want to accomplish? Could you provide a live example?

I think the best person to answer would be @mtogstad, as he did the change and tested it locally.

punnone:
@palashkaria I'm facing the same problem. Did you get a solution?

Heedster:
Fixed it. 2 things have to be done.
1. Adding a seLatLng to the tile [CanvasMap.js]
```
tiles[key] = {
        ...tile2LatLng({ x, y }, zoom),
        seLatLng: { ...tile2LatLng({ x: x + 1, y: y + 1 }, zoom) },
        x,
        y,
        markers: []
   };
```
2. Remove explicit width from the canvas tile [CanvasTile.js]
```
const canvasStyle = {
  width: "100%",
  height: "100%"
};
```

Updated the codesanbox for reference: https://codesandbox.io/s/6v5vx7z65w

PR #594, opened 2018-05-31
Add guard around mapDom event listener

bignimbus:
Fixes an issue where, even if the end user has a proper google maps API
mock, the GoogleMaps component will throw an error when in a test
env.  ReactDOM.findDOMNode(this.googleMapDom_) returns `null`.

This commit wraps the subsequent addEventListener call in an `if` block,
ensuring that an empty DOM doesn't add complications to test suites

Thank you for your consideration and your hard work on this great library!


itsmichaeldiego:
This is great @bignimbus ! Thanks, someone opened a similar PR but provided not as good description as yours. Merging and it will be present in the next release together with other group of changes as this one is not a blocker. Stay put.

PR #601, opened 2018-06-20
Update API.md

vulong23:
Because the parent node of marker has size 0x0, top 50% and right 50% would affect nothing. You only need to translate(-50%, -50%) to move it to the right position.

### DIFF
--- a/API.md
+++ b/API.md
@@ -240,8 +240,6 @@ Example (centering the marker):
 ```javascript
 const greatPlaceStyle = {
   position: 'absolute',
-  top: '50%';
-  left: '50%';
   transform: 'translate(-50%, -50%)';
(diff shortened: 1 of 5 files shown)


itsmichaeldiego:
@vulong23 Thanks for your contribution, you're right! 

The build is not passing as it was broken, its fixed now, would you mind rebasing with latest master?

itsmichaeldiego:
@vulong23 Mm the rebase did not work as expected, you've changed 5 files.

vulong23:
@itsmichaeldiego lol I messed up with git again, I'll create a new pull request

itsmichaeldiego:
@vulong23 No worries!

PR #615, opened 2018-07-16
Add prop `onTilesLoaded`

jonathanweiss:
There are [some events](https://developers.google.com/maps/documentation/javascript/events) that are fired by the Google Maps JavaScript API, but there are not available on the React component. With this PR, I'm passing through the event `tilesloaded`.

Please let me know what you think about this. I'd love to see my changes upstream.

### DIFF
--- a/API.md
+++ b/API.md
@@ -139,6 +139,7 @@ When the user changes the map type (HYBRID, ROADMAP, SATELLITE, TERRAIN) this fi
 Directly access the maps API - *use at your own risk!*
 
 #### onTilesLoaded (func)
+This function is called when the visible tiles have finished loading.
 
 ```javascript
 <GoogleMap  onGoogleApiLoaded={({map, maps}) => console.log(map, maps)} />
--- a/API.md
+++ b/API.md
@@ -138,6 +138,8 @@ When the user changes the map type (HYBRID, ROADMAP, SATELLITE, TERRAIN) this fi
 #### onGoogleApiLoaded (func)
 Directly access the maps API - *use at your own risk!*
 
+#### onTilesLoaded (func)
+
 ```javascript
 <GoogleMap  onGoogleApiLoaded={({map, maps}) => console.log(map, maps)} />
 ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,5 @@
 Add [google-map-clustering-example](https://github.com/istarkov/google-map-clustering-example)
 
-###[Unreleased]
 Add prop `onTilesLoaded` to react on the `tilesloaded` event
 
 ###0.9v
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 Add [google-map-clustering-example](https://github.com/istarkov/google-map-clustering-example)
 
-###1.1.0v
+###[Unreleased]
 Add prop `onTilesLoaded` to react on the `tilesloaded` event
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "google-map-react",
-  "version": "1.0.5",
+  "name": "@jfw/google-map-react",
```
itsmichaeldiego: Lets remove the @jfw

package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "1.0.6",
```
itsmichaeldiego: Lets not make a version from this PR, version will be bumped when necessary with one or multiple PRs

API.md
```diff
@@ -138,6 +138,8 @@ When the user changes the map type (HYBRID, ROADMAP, SATELLITE, TERRAIN) this fi
 #### onGoogleApiLoaded (func)
 Directly access the maps API - *use at your own risk!*
 
+#### onTilesLoaded (func)
+
```
itsmichaeldiego: What about the info?

package.json
jonathanweiss: Sorry for that. I needed to publish the component in order to use it in a project. I'll remove it on `master`.

package.json
jonathanweiss: Same here.

jonathanweiss:
I've applied all changes, please have another look @itsmichaeldiego.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 Add [google-map-clustering-example](https://github.com/istarkov/google-map-clustering-example)
 
+###[Unreleased]
```
itsmichaeldiego: @jonathanweiss Can we move this section to the bottom? And also, no need of add `Unreleased` as I am going to release it very soon after merged.

itsmichaeldiego:
Can you delete the latest commit? Only the maintainers set the version releases

src/google_map.js
```diff
@@ -663,6 +662,10 @@ export default class GoogleMap extends Component {
           this.heatmap.setMap(map);
         }
 
+        maps.event.addListener(map, 'tilesloaded', () => {
```
itsmichaeldiego: Should we add the listener all the times, or only when the user has set an onTilesLoaded prop? What do you think?

itsmichaeldiego:
@jonathanweiss Changes are looking great now! Thank you, I left a question for you before we merge this.

jonathanweiss:
Now we don't need `_onTilesLoaded()` any longer. 😉 Do you want to keep it for consistency?

itsmichaeldiego:
@jonathanweiss What do you need that we don't need it any longer?

jonathanweiss:
Well, all that it does is checking if `this.props.onTilesLoaded` is truthy before calling it. Since we're now only attaching the listener when that prop is truthy, we could directly call `this.props.onTilesLoaded();` in line 667.

itsmichaeldiego:
@jonathanweiss This is fine like this, thanks though!

jsharland:
Can someone please update the typescript definition to match as I could really do with the onTilesLoaded event.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google-map-react/index.d.ts

vertic4l:
@jsharland Type Definition are now up2date.
@jonathanweiss @itsmichaeldiego thank you!

PR #623, opened 2018-08-06
Updating map center when props changes during map initialization

cinek260:
When `center` prop changed during map initialization, old coordinates from old `center` props was active. In this PR I've added code to 
1. Store new props passed to component during map initialization.
2. Use newest center prop when map is rendered first time.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -507,7 +507,7 @@ export default class GoogleMap extends Component {
         }
 
         const centerLatLng = this.pendingProps_
-          ? this.pendingProps_.center
+          ? latLng2Obj(this.pendingProps_.center)
           : this.geoService_.getCenter();
 
         const propsOptions = {
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -195,6 +195,8 @@ export default class GoogleMap extends Component {
 
     this.googleMapDom_ = null;
 
+    this.pendingProps_ = null;
+
     if (process.env.NODE_ENV !== 'production') {
       if (this.props.apiKey) {
         console.warn(
@@ -373,6 +375,8 @@ export default class GoogleMap extends Component {
         });
         this._setLayers(nextProps.layerTypes);
       }
+    } else {
+      this.pendingProps_ = nextProps;
     }
   }
 
@@ -502,7 +506,9 @@ export default class GoogleMap extends Component {
           return;
         }
 
-        const centerLatLng = this.geoService_.getCenter();
+        const centerLatLng = this.pendingProps_
+          ? this.pendingProps_.center
+          : this.geoService_.getCenter();
 
         const propsOptions = {
           zoom: this.props.zoom || this.props.defaultZoom,


itsmichaeldiego:
@cinek260 This seems that can be done checking the props before rendering the map itself, would you explain the scenario where the center props can be changed before the map is rendered? 🤔

itsmichaeldiego:
@cinek260 Looking forward for your answer.

cinek260:
@itsmichaeldiego Sorry for late response. Here's a scenario:
I have a page where I am using google-map-react to display results from algolia search. On the center of the map I want to display first result from search. So I need to load map and results. I want to do it asynchronous, to have better page loading time. When `this.props.googleMapLoader` promise is waiting for being resolved, I get my search results and update `center` prop. Unfortunately map renders with center different from defined in `center` prop. This PR fixes it.

itsmichaeldiego:
@cinek260 My question is: why you don't wait for that to update before rendering the map?

itsmichaeldiego:
@cinek260 Great, thanks for this! I need some time to test it before merging it, but I will let you know as soon as I can and I will release it for next version.

JerryLeeCS:
Does this solve

`GoogleMap: defaultCenter prop changed. You can't change default props.` ?

itsmichaeldiego:
@cinek260 I did not forget about this, just did not have time to test it thoroughly . If you have the chance to send me a test-video that would be amazing.

cinek260:
@itsmichaeldiego ok, I will try to do it this week

itsmichaeldiego:
@cinek260 Thanks bro, sorry for that, I've been wrapped up in a couple of projects

cinek260:
@itsmichaeldiego Very sorry for being not responsive, I just couldn't find time for this 😞 In the meantime I found that my change was caused errors in some cases, so I've added another commit to fix it.

itsmichaeldiego:
@cinek260 That is great, no problem man, I am just looking forward to merge this as soon as you do it!

TiBarification:
Will it be fixed? I have same problem.

itsmichaeldiego:
@TiBarification I am waiting on @cinek260 to update the docs in order to merge this

src/google_map.js
```diff
@@ -502,7 +506,9 @@ export default class GoogleMap extends Component {
           return;
         }
 
-        const centerLatLng = this.geoService_.getCenter();
+        const centerLatLng = this.pendingProps_
+          ? latLng2Obj(this.pendingProps_.center)
+          : this.geoService_.getCenter();
```
pieplu: Your assign pendingProps_ but never "unassign" it (on line 379). `this.geoService_.getCenter();` will never be called after the map was initialized ( `this.map_ == true`). It's look's like weird, no?

pieplu:
I had a similar issue, my "center" props is defined durring the google map componant loading. When the map is finaly loaded `componantWillReciveProps` will not update the center, because the prevent check think the new props center is already setted...

@cinek260  It's looks like the bug is on this  [double if check](https://github.com/google-map-react/google-map-react/blob/master/src/google_map.js#L326)
```
        if (
          !currCenter ||
          Math.abs(nextPropsCenter.lat - currCenter.lat) +
            Math.abs(nextPropsCenter.lng - currCenter.lng) >
            kEPS
        ) {
          if (
            Math.abs(nextPropsCenter.lat - centerLatLng.lat) +
              Math.abs(nextPropsCenter.lng - centerLatLng.lng) >
            kEPS
          ) {
            this.map_.panTo({
              lat: nextPropsCenter.lat,
              lng: nextPropsCenter.lng,
            });
          }
        }
      }
```
The `currCenter` is not the real actual currentCenter. I think merging the 2 if, with a `||` will fix it.

Anyway, @TiBarification I found a workaround to fix it.
I save the center value on state, and update it, only after the `onGoogleApiLoaded` props will be called.

It's looks like this:

```javascript
state = {
  mapApiLoaded: false,
  center: {lat: 0, lng: 0}
};

componentDidUpdate(prevProps:Props) {
  if (this.state.mapApiLoaded && /*...*/ ) {
    // Do setState on center
  }
}


... (render)
<GoogleMap
  center={this.state.center}
  onGoogleApiLoaded={() => this.setState({
     mapApiLoaded: true,
     center: his.props.center
   })}
/>

```

Hope this help!

itsmichaeldiego:
@pieplu That's an excellent work around.

itsmichaeldiego:
@pieplu That's an excellent work around.

@TiBarification Could you check it works on your end?

TiBarification:
That's actually works, seems, that there no reason to modify code of library, we can use states directly to the center.

TiBarification:
```
I save the center value on state, and update it, only after the onGoogleApiLoaded props will be called.
```
It should be documented I think

Issue #625, opened 2018-08-07
Passive event listeners


johnhult:
Not sure if it's this library but according to [https://issuetracker.google.com/issues/63211698](url) Google Map should not give warnings when doing a Lighthouse Audit for not using passive eventListeners. Is this something in regards to this library? Doing a lighthouse audit I get the following error:

<img width="694" alt="screen shot 2018-08-07 at 14 22 59" src="https://user-images.githubusercontent.com/1528648/43775549-7292342c-9a4d-11e8-943d-f69463acf05e.png">

If it indeed isn't this library then feel free to close this issue.

itsmichaeldiego:
@johnhult Thanks for your help. It seems that its the google map api itself that its throwing an error, you can see the URL of the warning is actually the url of googleapis and not your URL, if it would be our library it will be showing the URL of your current project.

johnhult:
Yeah I thought it was kinda weird. What version of google maps is this library using? As mentioned in my comment version 3.32 of google maps should apparently not give these errors. But this uses a higher version than that, right?

I'll try and look into it a bit more aswell but just want to make sure I have the correct assumptions first 😄

itsmichaeldiego:
@johnhult Our library uses latest released version always, you can specify the version in `bootstrapURLKeys` doing like: 

```
<GoogleMapReact
  ...
  bootstrapURLKeys={{
    ...
    v: '3.33',
  }}
/>
```

itsmichaeldiego:
@johnhult @jooj123 Fix #631 has been merged and released in `1.0.7`

Let me know if it works.

PR #631, opened 2018-08-15
Add passive scroll

jooj123:
Added passive scroll event (which is more performant).

See: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener (`passive` in options)

Needed the feature detection for browsers that dont support it

**Note:** hard to test this as the demos for `google-map-react` are all busted due to incorrect google maps key

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -26,6 +26,7 @@ import shallowEqual from './utils/shallowEqual';
 import isPlainObject from './utils/isPlainObject';
 import isArraysEqualEps from './utils/isArraysEqualEps';
 import detectElementResize from './utils/detectElementResize';
+import addPassiveEventListener from './utils/passiveEvents';
 
 // consts
 const kEPS = 0.00001;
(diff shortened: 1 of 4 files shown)


itsmichaeldiego:
@jooj123 Is this related to #625 ?

jooj123:
@itsmichaeldiego this is the only place in this lib where it needs a passive event - so it should get rid of the warnings in the console and things should be a bit smoother

jooj123:
@itsmichaeldiego updated. Not sure how i can unit test this as its browser specific functionality 
I cant even validate it working correctly without the demo working incorrect due to google maps key

itsmichaeldiego:
@jooj123 That is fine man! Thanks for this! Could you provide something like a video or whatever so I can see this is working? I really want to merge this but I've no time to test it :(

itsmichaeldiego:
@jooj123 Bump!

jooj123:
@itsmichaeldiego sorry havent had a chance to test this one - will do sometime over the weekend or next week

itsmichaeldiego:
@jooj123 What do you mean that the internal demo worked? We're working on new examples at https://github.com/google-map-react/google-map-react-examples

PR #634, opened 2018-08-27
Custom div style options

### LINKED ISSUE #632: Custom overlayview style attributes
Would be great if it would be possible to add custom style properties to div or panes.

https://github.com/google-map-react/google-map-react/blob/master/src/google_map.js#L589

In my case I need to add `pointer-events: none` (an overlapping issue with custom tiles and an interesting way of rendering stuff in general).

Could that be possible? I can get there trough refs, but i would prefer an API.

jurgenzz:
fixes #632

### DIFF
--- a/API.md
+++ b/API.md
@@ -153,6 +153,10 @@ To prevent warning message add _yesIWantToUseGoogleMapApiInternals_ property to 
  />
  ```
 
+#### overlayViewDivStyle (object)
+
+Set custom style options on `div` (marker container element) created by OverlayView, for example: `{pointerEvents: 'none'}`.
+
 ## Child Component API
 
 ### parameters
--- a/API.md
+++ b/API.md
@@ -155,7 +155,7 @@ To prevent warning message add _yesIWantToUseGoogleMapApiInternals_ property to 
 
 #### overlayViewDivStyle (object)
 
-Set custom style options on `div` (marker container element) created by OverlayView, for example: `{pointerEvents: 'none'}`.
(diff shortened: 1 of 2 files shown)


jurgenzz:
Usage: 
```

jurgenzz:
I'd guess no, as for name `wrapper` i would think of the `div` surrounding `<GoogleMapReact />` component itself, not just marker, which actually goes into a pane. A lot of other overlays there :) . Then we could think of `markerContainerStyle` and other, but by just having a quick look at source, you'd figure out that `this.div` is probably getting all the styles. But i can change to something else tomorrow :)

itsmichaeldiego:
@jurgenzz Yes, I think we could find a more descriptive name for this, as this div can be any div.

itsmichaeldiego:
@jurgenzz How are we looking on this? I would like to merge it and release it so we can close #632

itsmichaeldiego:
@jurgenzz That is great! Thanks, I just realize something (sorry I didn't before) but we need to add this to the DOCS so other people know is there, more specifically in the [API.md file](https://github.com/google-map-react/google-map-react/blob/master/API.md)

jurgenzz:
If you want, you can release it as it is for now. I've created reminder for Monday, when i'll be at work i'll update docs and API.md file.

itsmichaeldiego:
@jurgenzz We can wait, thanks!

jurgenzz:
Ye, sorry, updated now :)

itsmichaeldiego:
@jurgenzz Great work man! I am releasing this in `1.0.9`!

itsmichaeldiego:
@jurgenzz Released in 1.0.9!

PR #643, opened 2018-09-18
Use React 16 portal to render map overlay

### LINKED ISSUE #642: Context (new API) is not propagated to google-map-react children

**Expected Behaviour**

**Current Behaviour**

clayne11:
This allows the new context API to propagate context properly.

Fixes #642.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -239,7 +239,7 @@ export default class GoogleMap extends Component {
     this.zoomAnimationInProgress_ = false;
 
     this.state = {
-      overlayCreated: false,
+      div: null,
     };
   }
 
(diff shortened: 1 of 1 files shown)


src/google_map.js
```diff
@@ -239,7 +239,7 @@ export default class GoogleMap extends Component {
     this.zoomAnimationInProgress_ = false;
 
     this.state = {
-      overlayCreated: false,
+      div: null,
```
itsmichaeldiego: Can we rename `div` to `overlay`?

clayne11:
Done.

stuartherring:
This change breaks the component when used with React versions prior to 16.

itsmichaeldiego:
@stuartherring @abumusamq Sorry for the inconvenience! This was totally my bad, I just fixed this and released in `1.0.8`.

@clayne11 do you mind opening the PR again? We will add this to the next major release.

Or, I think there should be a way to identify if the user has React 16 or not and use createPortal if so, but that would be a bit more tricky.

clayne11:
No problem I can do that. Are you planning to cut a new major release with this soon? React 16 has been out for a year now (released Sept 26, 2017) so I think it's about time to move forward.

itsmichaeldiego:
@clayne11 Yes, we would like to do that soon but lately it has been only me taking care of the library, any help would be accepted lovely xD.

PR #647, opened 2018-09-21
Use React 16 portal to render map overlay

### LINKED ISSUE #642: Context (new API) is not propagated to google-map-react children

clayne11:
This allows the new context API to propagate context properly.

Fixes #642.

Note: This is a breaking change since it will no longer support React < 16 and should be released as a new major version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,8 @@
   "homepage": "https://github.com/google-map-react/google-map-react#readme",
   "peerDependencies": {
     "prop-types": "^15.5.6",
-    "react": "^0.14.0 || ^15.0.0 || ^16.0.0",
-    "react-dom": "^0.14.0 || ^15.0.0 || ^16.0.0"
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0"
   },
   "dependencies": {
     "@mapbox/point-geometry": "^0.1.0",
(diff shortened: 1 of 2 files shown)


jooj123:
this effectively means this component no longer supports react 15 - is this wise? we still use this component in a react 15 code base - its not always that trivial to move to the latest react on a large codebase

itsmichaeldiego:
@jooj123 I am wondering if we should catch when the user is using 16+ and apply this code when so.

jooj123:
@itsmichaeldiego yeah some sort of fallback would be good 
maybe even a 3rd party one like: https://github.com/tajo/react-portal

clayne11:
React 16 has been out for over a year at this point and upgrading really isn't that difficult IMO. I can understand that we don't want to making a breaking change like this for in minor version but we can't continue to support React < 16 indefinitely.

It seems reasonable to me, at this point, that we publish a new major version with this breaking change and that the current version continues to live on as is.

If you want the latest and greatest features of anything you have to upgrade versions and deal with the migration (which from React 15 - 16 really is not that much). If you're happy with React 15 then by all means you can keep using it but the community has to move forward. A year is a long time in the frontend JavaScript world.

itsmichaeldiego:
@clayne11 We added this with backwards compatibility in #696, so this becomes obsolete, thanks.

PR #656, opened 2018-10-11
Pass map instance to onDrag handler

bogomya:
Goal:
Add possibility to work with google map api in the onDrag handler.
+ documentation improving

### DIFF
--- a/API.md
+++ b/API.md
@@ -126,6 +126,8 @@ Default: false
 
 #### onZoomAnimationStart (func)
 
+#### onDrag ((map) => void)
+
 #### onZoomAnimationEnd (func)
 
 #### onMapTypeIdChange (func)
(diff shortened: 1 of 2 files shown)


itsmichaeldiego:
@bogomya Could you explain how this improves and provide an example? Just for documentation, I totally understand what you are doing

bogomya:
The idea was to get map bounds on drag event to load new markers. I could use onChange, but onChange is called also when zoom event is occurred, and also on initial load.

itsmichaeldiego:
@bogomya Thank you so much! This would be helpful. Could you provide a live example of this working? I need to check its working before releasing it.

itsmichaeldiego:
@bogomya Merged and deployed in version `1.1.0`, please try it and let me know

Issue #673, opened 2018-11-12
v1.1.0 Not adding libraries (drawing/places etc) to maps api object


tim-soft:
Upgrading from v1.0.9 to v1.1.0, libraries are no longer added to the maps api object. 

```js
<GoogleMapReact
  bootstrapURLKeys={{
    key: 'xxxxxxxxxxxxxx',
    libraries: ['places', 'drawing'],
  }}
  yesIWantToUseGoogleMapApiInternals
  onGoogleApiLoaded={({ maps }) => {
    console.log(maps.places); // undefined
    console.log(maps.drawing); // undefined;
  }}
/>
```

itsmichaeldiego:
@tim-soft Good catch!

This was introduced in https://github.com/google-map-react/google-map-react/pull/593

I think for the time being, instead of passing `libraries` as you were, you can do:
```
<GoogleMapReact
  bootstrapURLKeys={{
    key: 'xxxxxxxxxxxxxx',
    libraries: ['places', 'drawing'],
  }}
  drawingLibrary={true}
  placesLibrary={true}
  yesIWantToUseGoogleMapApiInternals
  onGoogleApiLoaded={({ maps }) => {
    console.log(maps.places); // undefined
    console.log(maps.drawing); // undefined;
  }}
/>
```

itsmichaeldiego:
@DonovanDeSmedt Mind giving this a check? I think we should change the approach to use heatmapLibrary the same way we used to add the libraries in the URL:

`
<GoogleMapReact
  bootstrapURLKeys={{
    key: 'xxxxxxxxxxxxxx',
    libraries: ['places', 'drawing', 'heatmap']
  }}
  yesIWantToUseGoogleMapApiInternals
  onGoogleApiLoaded={({ maps }) => {
    console.log(maps.places);
    console.log(maps.drawing);
  }}
/>
`

itsmichaeldiego:
@tim-soft This was fixed and deployed in `1.1.1`, please let me know!

faeron:
are both ways (as described in #681) broken now? For me using the "old" way doesn't work as long as `heatmapLibrary={true}` is used as described in the docs. It overrides set libraries [here](https://github.com/google-map-react/google-map-react/blob/599195510e596173f94cc1feca49e2eae700fe6a/src/loaders/google_map_loader.js#L74)

The "new" way with using props for each library isn't working at all, as only the heatmapLayer prop seems to be usable, other libraries are not implemented yet I assume?

[...]

PR #686, opened 2018-11-26
Update readme

maruf89:
Update readme to show how to add Google Map API Libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,19 @@ There is no need to place a `<script src=` tag at top of page. The Google Maps A
 
 Now every object on the map can be hovered (however, you can still use css hover selectors if you want). If you try zooming out here [example](http://google-map-react.github.io/google-map-react/map/main), you will still be able to hover on almost every map marker.
 
+### API Library Support
+
+Request Google Map libraries by adding any of the following as props to your <GoogleMapReact /> object.
+
+```
+<GoogleMapReact
+  heatmapLibrary
+  geometryLibrary
+  placesLibrary
+  drawingLibrary
+  bootstrapURLKeys=...
+```
+
 ## Examples
 
 * Placing react components on the map:


barakplasma:
I think this PR should wait for https://github.com/google-map-react/google-map-react/issues/673 to be fixed first, since adding any library along with the heatmap layer currently causes the other libraries not to be loaded.

itsmichaeldiego:
@maruf89 Ideally we want to pass them on the libraries object instead of one by one like that. CC @barakplasma

itsmichaeldiego:
@barakplasma Thoughts on my last comment?

barakplasma:
> @barakplasma Thoughts on my last comment?

I don't have a strong opinion on which API is better, whether to have a list of boolean props for each library or one 'libraries' object with the same props. I am only mildly inclined to 'namespace' the props under the libraries object, since it does seem a bit more neatly organized than the alternative.

itsmichaeldiego:
@barakplasma That's a good idea! Would you like to work on that? Or @maruf89 ?

barakplasma:
@maruf89 would you like to work on this?
I won't be able to work on this any time soon.

maruf89:
No, I'm not doing anything with computers or coding anymore for a long long
while

wt., 16.04.2019, 12:23 użytkownik Michael Salaverry <
notifications@github.com> napisał:

itsmichaeldiego:
@barakplasma Let me know if you still interested on this othwerwise I will create an issue and ask for help

barakplasma:
Just to clarify, we are talking about fixing #673, or are we talking about updating the readme? I think we should recommend one _"right"_ way to add libraries.

itsmichaeldiego:
@barakplasma I am talking about recommending one "right" way to add libraries, which I think will be under `libraries: ['bar', 'foo']`. And keeping this way to pass libraries instead of two.

barakplasma:
I agree with putting libraries to load in the library object.

itsmichaeldiego:
@barakplasma If so I think we should remove the other way, so there is only one way and no collision

itsmichaeldiego:
Seems that this was taken care by #921

itsmichaeldiego:
@barakplasma @maruf89 Let me know if that works for you folks

PR #696, opened 2018-12-18
Added React 16 createPortal with backwards compatibility

### LINKED ISSUE #642: Context (new API) is not propagated to google-map-react children

c58:
Fixes #642 

Same as https://github.com/google-map-react/google-map-react/pull/647 but still works with React 15.

NOTE: no breaking changes here, could be released as a minor version.


itsmichaeldiego:
@c58 Thank you so much! Merging this!

c58:
@itsmichaeldiego thank you for the fast feedback 👍

PR #722, opened 2019-02-15
Add shouldUnregisterMapOnUnmount prop to persist map object on unmount

KevinRamsunder:
Each call to initialize a new map incurs an additional cost with Google's new billing platform.

In order to re-use the map instance between individual components, the following changes have been made:

1) Adds a `shouldUnregisterMapOnUnmount` prop to prevent the deletion of the map object, so it can be re-used.

2) Exposes the ref to the `googleMapDom_` via the `_onGoogleApiLoaded` callback, so it can be cached in a consuming application.

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -137,6 +137,7 @@ export default class GoogleMap extends Component {
     style: PropTypes.any,
     resetBoundsOnResize: PropTypes.bool,
     layerTypes: PropTypes.arrayOf(PropTypes.string), // ['TransitLayer', 'TrafficLayer']
+    shouldUnregisterMapOnUnmount: PropTypes.bool,
   };
 
   static defaultProps = {
@@ -161,6 +162,7 @@ export default class GoogleMap extends Component {
     layerTypes: [],
     heatmap: {},
     heatmapLibrary: false,
+    shouldUnregisterMapOnUnmount: true,
   };
 
   static googleMapLoader = googleMapLoader; // eslint-disable-line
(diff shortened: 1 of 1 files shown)


KevinRamsunder:
Fixed tests. Attached a screenshot of this fix working in my app (unfortunately, I can't share more than this due to proprietary reasons). Each "recordevent" is equal to one billing event.

The image on the left was before this fix, and the image on the right is after the fix. Note, the caching of the map/ref objects must be done in the consuming application -- this PR simply exposes the ref to facilitate that process.

screen shot 2019-02-18 at 6 58 40 pm

itsmichaeldiego:
@KevinRamsunder This is looking good! And will decrease the billing of the users, thanks for the change! I am deploying it soon

KevinRamsunder:
@itsmichaeldiego thanks! any idea when this might be tagged for release?

itsmichaeldiego:
@KevinRamsunder Published in v 1.1.3

Let me know if everything is working as expected, thanks.

itsmichaeldiego:
@KevinRamsunder This PR caused 
- https://github.com/google-map-react/google-map-react/issues/736
- https://github.com/google-map-react/google-map-react/issues/737 
Can you take a look?

itsmichaeldiego:
@KevinRamsunder Reverting for now, cannot afford to have a broken version, please open a new PR that works properly! Thanks in advance

KevinRamsunder:
@itsmichaeldiego the issues you mentioned don't appear to be related to my change - it was simply a prop change which is coded with backwards compatibility in mind. No functionality changes if the user doesn't opt in. The issues you pointed out are likely related to https://github.com/google-map-react/google-map-react/pull/726, which explicitly states that it ` handles bootstrapURLKeys change and reloads the Google Maps API`

With that being said, I do understand why you reverted this - but hope we can get it back in the future. Thanks!

itsmichaeldiego:
@KevinRamsunder Lets create another PR and I will merge it!

goncharov:
Hi, @KevinRamsunder, @itsmichaeldiego. Can you please provide example of usage of this option? I don't get how to reuse map object between google-map-react component instances

RithikaChowta:
@KevinRamsunder @itsmichaeldiego Any update on the above query?

managervcf:
@KevinRamsunder Providing `shouldUnregisterMapOnUnmount={false}` doesn't unregister the map instance, therefore reduces number of map intances created by [Maps JavaScript API](https://developers.google.com/maps/documentation/javascript), is that correct? 


I'm looking to minimze the costs of Maps JavaScript API by reducing number of map loads - [Price per MAP LOAD](https://developers.google.com/maps/billing-and-pricing/pricing#dynamic-maps).

PR #726, opened 2019-03-06
Google Map loader respecting the app language change

tomwagner:
**Problem**: When map language is dependent on app settings, the map doesn't respects current language and region value passed in `bootstrapURLKeys` object. New loader handles  `bootstrapURLKeys` change and reloads the Google Maps API (because of lack support to change language in API).

You can try this new loader here: [https://codesandbox.io/s/pppvk1o9z0](https://codesandbox.io/s/pppvk1o9z0)

PR #728, opened 2019-03-07
Fix: Update heatmap layer when heatmap positions prop changes

barakplasma:
This PR is somewhere between a bug fix and a new feature.

Before this PR, if the heatmap positions prop was updated, the new data would not be displayed on the heatmap layer.
After this PR, if the heatmap.positions data is changed, then the heatmap layer will run a setData call and update the heatmap accordingly.
To demonstrate this, I updated the develop/GMapHeatmap.js example with a random data generator.

Along the way, I changed how the API key is set in the develop/ examples. I am happy to remove that change, but I think it's better this way. Additionally, I noticed that the API key in the develop folder doesn't work currently.

### DIFF
--- a/develop/GMap.js
+++ b/develop/GMap.js
@@ -18,6 +18,7 @@ import GoogleMapReact from '../src';
 
 import ptInBounds from './utils/ptInBounds';
 import withStateSelector from './utils/withStateSelector';
+import { GOOGLE_API_KEY } from './config/Google_API_key';
 
 export const gMap = (
   {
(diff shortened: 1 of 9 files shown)


barakplasma:
How it looks
------------
giphy

barakplasma:
@itsmichaeldiego please send me some feedback on if this change is acceptable to you. I don't mind changing the example back to a static heatmap, and perhaps adding a separate animated heatmap example.

barakplasma:
@itsmichaeldiego any news on this PR?

barakplasma:
Thanks for the approval @itsmichaeldiego .
Can we merge this and bump the version number to 1.1.3 ?

itsmichaeldiego:
@barakplasma Published in v 1.1.3

Let me know if everything is working as expected, thanks.

epalmans:
image
v1.1.3 causes issues on SSR build (Gatsby in this example)

epalmans:
PS. Seems NOT to be related to this PR. But to this one instead: https://github.com/google-map-react/google-map-react/pull/726

PR #735, opened 2019-04-02
Fix isomorphic rendering

caresx:
On the server, `window` is undefined. Importing this file (which is done when just `import`ing google-map-react) sets a callback on `window` (undefined), crashing the server.


itsmichaeldiego:
@caresx Does onGoogleMapAPILoaded work for you though?

itsmichaeldiego:
@caresx Thanks, we will roll back changes and fix this later.

PR #742, opened 2019-04-10
Pass zoom value to zoom animation callbacks

eneo5541:
Pretty minor quibble, but it would be helpful to get the map's new zoom level returned on the zoom animation callbacks. 
Especially since they are already set up to pass params to the callback prop, but nothing is passed.


itsmichaeldiego:
@eneo5541 Sorry I just seee this, it seems that a nice change, but are you actually using this?

poptocrack:
@itsmichaeldiego I do use it. This is a pretty nice addition actually.

itsmichaeldiego:
@poptocrack Could you create an example for users to see that they can use that?

PR #754, opened 2019-05-16
Add prop 'onDragEnd'

eneo5541:
There is no callback dragEnd, which I feel would be a helpful callback to have.

The `dragEnd` event fired by the maps API fires as soon as the user releases the mouse, while the map inertia continues to pan the map. This callback therefore waits until the `idle` event is fired, then cleans up the event listener and invokes `props.dragEnd`

Please let me know if there's a better way to do this!

### DIFF
--- a/API.md
+++ b/API.md
@@ -128,6 +128,9 @@ Default: false
 
 #### onDrag ((map) => void)
 
+#### onDragEnd ((map) => void)
+When the map stops moving after the user drags. Takes into account drag inertia.
+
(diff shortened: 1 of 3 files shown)


itsmichaeldiego:
@eneo5541 Merged and released in 1.1.5, let me know if it works

itsmichaeldiego:
@eneo5541 @eldyvoon Merged and released in 1.1.5, let me know if it works

PR #773, opened 2019-07-30
Change yesIWantToUseGoogleMapApiInternals

tanohzana:
I think `yesIWantToUseGoogleMapApiInternals` variable name is long enough, and we should stick to what makes more sense to the whole community. `useGoogleMapApiInternals` is more understandable as a boolean. Let me know what you think.

Other than that, really cool package 🙂

### DIFF
--- a/API.md
+++ b/API.md
@@ -147,11 +147,11 @@ This function is called when the visible tiles have finished loading.
 <GoogleMap  onGoogleApiLoaded={({map, maps}) => console.log(map, maps)} />
 ```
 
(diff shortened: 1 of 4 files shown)


darbio:
I agree, however it should be kept for backwards compatibility and marked as deprecated (to be removed in the next breaking change), rather than removed altogether.

itsmichaeldiego:
Idem with @darbio: @tanohzana we should do backwards compatibility here

PR #790, opened 2019-09-27
Fixed issue with `undefined` initial zoom

More4ever:
Resolved Issue: 
mount zoom value is not set, in process of work we set zoom to any value (eg. 12). 

`Math.abs(nextProps.zoom - this.props.zoom) > 0` condition is always `false`

`Math.abs(12 - undefined)` = `NaN`

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -344,12 +344,10 @@ export default class GoogleMap extends Component {
           }
         }
       }
-
-      if (!isEmpty(nextProps.zoom)) {
-        // if zoom chaged by user
-        if (Math.abs(nextProps.zoom - this.props.zoom) > 0) {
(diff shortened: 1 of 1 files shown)


itsmichaeldiego:
@More4ever Is this from an open issue?

More4ever:
Not sure it was too many days ago

Probably it exists only as PR. I can create issue with same description and source code links if it is required

itsmichaeldiego:
@More4ever The problem is that I don't see the issue happening, so I don't know what you are fixing. Therefore yes, please!

itsmichaeldiego:
@More4ever I think this is a duplicate of https://github.com/google-map-react/google-map-react/pull/864, and I think #864 fixes the problem way better, what do you think?

PR #793, opened 2019-10-03
Extend `onGoogleApiLoaded` documentation

jorrit:
Documents how a load failure of the Google API affects the `onGoogleApiLoaded` callback.


itsmichaeldiego:
I think this is obvious, thanks though.

Issue #677, opened 2018-11-20
X of undefined on mouse enter / move on map


Umbrellagun:
Using v1.1.0

Whenever I include even the simplest marker on the map I get this error when the mouse enters the map:

`Uncaught TypeError: Cannot read property 'x' of undefined`

I'm not sure if it's actually directly breaking anything because the markers still show up, but I could see it causing other wierd behavoir maybe. I'm 'assuming' that this error is related to the "Internal Hover Algorithm". But this is just a guess.

A-Tokyo:
@Umbrellagun I have the same issue, Did you figure out a work around maybe? 🤔 Thanks!

Umbrellagun:
@A-Tokyo Not yet, no. I'll let you know if I get to it though!

A-Tokyo:
@Umbrellagun I am giving it some time of investigation now.

Weirdly when I updated:

```js
GoogleMap.defaultProps = {
  distanceToMouse: function distanceToMouse(pt, mousePos /* , markerProps */) {
    return Math.sqrt((pt.x - mousePos.x) * (pt.x - mousePos.x) + (pt.y - mousePos.y) * (pt.y - mousePos.y));
  },
```
to

```js
GoogleMap.defaultProps = {
  distanceToMouse: function distanceToMouse(pt, mousePos /* , markerProps */) {
    console.log('pt', pt);
    console.log('mousePos', mousePos);
    return Math.sqrt((pt.x - mousePos.x) * (pt.x - mousePos.x) + (pt.y - mousePos.y) * (pt.y - mousePos.y));
  },
```

It worked perfectly, no erros were thrown and the values were logged. I am starting to think it's related to Async React, Are you using Async React as well?

itsmichaeldiego:
Could you guys upload a live-example?

Umbrellagun:
@A-Tokyo I'm not using Async React, no. I'll take a look at the default props though, I don't think I have anything in the code for this at all currently. All you did was add console logs, right? You think it was just the time delay that it added that made it stop throwing the error?

JoshuaDoshua:
Anybody find a resolution for this?

A-Tokyo:
@Umbrellagun Yes exactly, I think the time delay that it added made it stop throwing the error.

Umbrellagun:
[...]

PR #849, opened 2020-03-29
Added empty object check GoogleMapMarkers

kumiau:
Issue #677


itsmichaeldiego:
@kumiau This seems to be taken care in https://github.com/google-map-react/google-map-react/pull/873#pullrequestreview-452862273, let me know if I am wrong.

PR #864, opened 2020-05-05
On resize it's getting undefined minZoom

AfzalH:
`this.props.options.minZoom` is not getting the value that's been passed (on window resize), instead getting `undefined` setting the minZoom level to default 3... should actually be `options.minZoom`

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -756,7 +756,7 @@ export default class GoogleMap extends Component {
           if (this.resetSizeOnIdle_) {
             this._setViewSize();
             const currMinZoom = this._computeMinZoom(
-              this.props.options.minZoom
+              options.minZoom
             );
 
             if (currMinZoom !== this.minZoom_) {
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -755,9 +755,7 @@ export default class GoogleMap extends Component {
         maps.event.addListener(map, 'idle', () => {
           if (this.resetSizeOnIdle_) {
             this._setViewSize();
-            const currMinZoom = this._computeMinZoom(
-              options.minZoom
-            );
+            const currMinZoom = this._computeMinZoom(options.minZoom);
 
             if (currMinZoom !== this.minZoom_) {
               this.minZoom_ = currMinZoom;


itsmichaeldiego:
@AfzalH When do you have this error? Could you reproduce it for me?

AfzalH:
> @AfzalH When do you have this error? Could you reproduce it for me?

There's no `error`, it's just getting the undefined value on window resize and resetting the minZoom to 3.
This is happening because in this case we are passing `this.props.options.minZoom` to the function `_computeMinZoom`. In other places we are passing `options.minZoom`.

To reproduce above each `_computeMinZoom` call put a console.log for the value being passed. Initialize a map with a minZoom option 1 and try to resize browser window. console.log would show 'undefined' and _computerMinZoom will set the minZoom to '3' which is the default value programmed to be set in case of missing minZoom (but in this case minZoom isn't missing, it's just not read properly)

lanre-mark:
You could also change the line to `const currMinZoom = this._computeMinZoom(
              this.props.options.minZoom || options.minZoom
            );` as a caution check.
I wondered the library worked in some circumstances though options.minZoom has always been the argument passed into the _computeMinZoom call elsewhere in the codebase.

AfzalH:
> You could also change the line to `const currMinZoom = this._computeMinZoom( this.props.options.minZoom || options.minZoom );` as a cautionary check.
> I wondered the library worked in some circumstances though options.minZoom has always been the argument passed into the _computeMinZoom call elsewhere in the codebase.

I also thought the same ... passing `this.props.options.minZoom || options.minZoom`
However, since I noticed on everywhere else it's options.minZoom , I assumed this.props.options.minZoom was a typo/mistake. That's why suggested to get rid of this.prop. in this PR

lanre-mark:
> > You could also change the line to `const currMinZoom = this._computeMinZoom( this.props.options.minZoom || options.minZoom );` as a cautionary check.
> > I wondered the library worked in some circumstances though options.minZoom has always been the argument passed into the _computeMinZoom call elsewhere in the codebase.
> 
> I also thought the same ... passing `this.props.options.minZoom || options.minZoom`
> However, since I noticed on everywhere else it's options.minZoom , I assumed this.props.options.minZoom was a typo/mistake. That's why suggested to get rid of this.prop. in this PR

you're right!

itsmichaeldiego:
@AfzalH This sounds right to me, I will merge and release this soon in a new version!

itsmichaeldiego:
@AfzalH @lanre-mark Released in `v2.0.1`, let me know if its working as expected

lanre-mark:
Thanks, works now.

PR #873, opened 2020-05-26
Moves GoogleMapMarkers MarkerDispatcher binds to componentDidMount

### LINKED ISSUE #843: Cannot read property 'x' of undefined error for basic example when moving mouse over map
Hi!

I'm getting the "Cannot read property 'x' of undefined" error message for a basic example created in the codesandbox. The error appears when I try to move a mouse cursor over a map.

Here's the link to the codesandbox: https://codesandbox.io/s/google-map-react-error-with-marker-9zwdm

To reproduce the error, remove/comment the `distanceToMouse` property from the map component.

The error message is the same as for #677 issue, but I'm not sure that it's cause by the same problem.

The implementation for the `distanceToMouse` in the codesandbox logs `pt` and `mousePos` values to the console. When moving mouse, a marker object is sometimes `undefined` which causes default implementation of the `distanceToMouse` to fail.

BTW, I've tried the same code in a project created with Create React App and everything works fine (react-scripts version is 3.4.0). 

I don't know why sometimes the `distanceToMouse` function receives `undefined` instead of an object for marker.

The quick workaround is to provide a custom implementation for the `distanceToMouse` function and check for `undefined`.

bdell:
Resolves #843 
This should also negate the need to merge [this PR](https://github.com/google-map-react/google-map-react/pull/849)

The issue above was caused because the `pt` argument in `distanceToMouse` was undefined. Tracing backwards I found out that value corresponded to the `this.dimensionsCache_[childKey]` value in `GoogleMapMarkers`, which when an erroring, was an empty object.

console logged values when erroring

Instead of inserting an empty object check into  `_onChangeHandler` in `GoogleMapMarkers` or overriding the `distanceToMouse` function on the child component, I went ahead and moved the `MarkerDispatcher` binds for `GoogleMapMarkers` into `componentDidMount`.

Previously, those binds were done in the constructor causing `dimensionsCache_` to be the default value (an empty object) instead of the updated value set inside of the `GoogleMapMarkers.render` function.

This change ensures that the `this.dimensionsCache_[childKey]` is defined when called in `_onChangeHandler` and thus the `pt` value is defined in the subsequent `distanceToMouse` call.

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -48,13 +48,6 @@ export default class GoogleMapMarkers extends Component {
 
   constructor(props) {
     super(props);
-    this.props.dispatcher.on('kON_CHANGE', this._onChangeHandler);
-    this.props.dispatcher.on(
-      'kON_MOUSE_POSITION_CHANGE',
-      this._onMouseChangeHandler
-    );
-    this.props.dispatcher.on('kON_CLICK', this._onChildClick);
-    this.props.dispatcher.on('kON_MDOWN', this._onChildMouseDown);
 
     this.dimensionsCache_ = {};
     this.hoverKey_ = null;
@@ -62,6 +55,16 @@ export default class GoogleMapMarkers extends Component {
     this.allowMouse_ = true;
 
     this.state = { ...this._getState(), hoverKey: null };
+  }
+
+  componentDidMount() {
+    this.props.dispatcher.on('kON_CHANGE', this._onChangeHandler);
+    this.props.dispatcher.on(
+      'kON_MOUSE_POSITION_CHANGE',
+      this._onMouseChangeHandler
+    );
+    this.props.dispatcher.on('kON_CLICK', this._onChildClick);
+    this.props.dispatcher.on('kON_MDOWN', this._onChildMouseDown);
   }
 
   shouldComponentUpdate(nextProps, nextState) {


sbichenko:
Is there any reason not to merge this?

itsmichaeldiego:
@sbichenko Nope, just time, would you like to become a contributor? We're looking ;)

bdell:
@itsmichaeldiego I wouldn't mind doing more fixes like this one if y'all are looking for contributors.  I'm start a new job in a couple weeks but after that I'm defs down to tackle some stuff.

itsmichaeldiego:
@bdell  That would be great, you can work in the issue of your preference, you can [filter by help wanted label](https://github.com/google-map-react/google-map-react/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22) also.

itsmichaeldiego:
@bdell Released in `2.0.1`, let me know if its working as you expect

bdell:
@itsmichaeldiego just checked and `2.0.1` seems to be working as intended.  Let me know if there is anything else I can help with, thanks!

itsmichaeldiego:
@bdell If you have a chance to take a look at https://github.com/google-map-react/google-map-react/issues/901 would be cool, I won't be able till tomorrow and it seems quite critical, let me know.

Also, we have a Slack if you're interested to join :+1:

karldanninger:
@itsmichaeldiego I want to join the Slack, pleeeaaassseee :)

itsmichaeldiego:
@karldanninger @bdell Here's an invite link for you folks https://join.slack.com/t/google-map-react/shared_invite/zt-g0ri7tth-SfilBwep9mOvKsOY5O4O3A

bdell:
@itsmichaeldiego sorry to not have gotten back to you, been super busy with stuff.  I can try and take a look at that issue after next weds if it still needs fixing.  Also I joined the slack so I’ll see ya around there :)

itsmichaeldiego:
@bdell We took care of it, but there is more stuff we can do, I've tagged all the important issues with the `help wanted` label, you can check it here https://github.com/google-map-react/google-map-react/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

PR #921, opened 2020-08-20
Added all google maps api libraries to api loader

### LINKED ISSUE #673: v1.1.0 Not adding libraries (drawing/places etc) to maps api object

### LINKED ISSUE #922: Add support to all libraries.
Current implementation only supports the `visualization` library (for some reason naming it as heatMapLibrary?). google maps has 4 libraries available: `visualization`, `geometry`, `drawing`, `places`.

It would be great to have support to all of them, so they are available on the `maps` object in the `onGoogleApiLoaded` method.

I propose a fix while still supporting the `heatMapLibrary` prop in this PR:
https://github.com/google-map-react/google-map-react/pull/921

aviyadeveloper:
Supports places, visualization, places, and geomerty libs.
keeps support for previous heatMapLibrary prop to avoid breaking older usage.

Closes #922
Closes #673

### DIFF
--- a/src/loaders/google_map_loader.js
+++ b/src/loaders/google_map_loader.js
@@ -13,7 +13,7 @@ const _customPromise = new Promise((resolve) => {
 });
 
 // TODO add libraries language and other map options
-export default (bootstrapURLKeys, heatmapLibrary, libraries) => {
+export default (bootstrapURLKeys, heatmapLibrary) => {
   if (!$script_) {
     $script_ = require('scriptjs'); // eslint-disable-line
   }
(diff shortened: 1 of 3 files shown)


itsmichaeldiego:
Closes https://github.com/google-map-react/google-map-react/issues/922

aviyadeveloper:
Oh I see now it is handled here:
```javascript
const params = Object.keys(bootstrapURLKeys).reduce(
  (r, key) => `${r}&${key}=${bootstrapURLKeys[key]}`,
  ''
);
```

But there are 3 issues with this implemntation:
1. There's no validation that libraries are correct
2. If user puts a library, and the `heatMapLibrary` prop is `True` the the query-params-string will be broken because of the next line:
```
const libraries = heatmapLibrary ? '&libraries=visualization' : '';
```
resulting in two `&libraries=`

3. I would add documentation about the libraries option.

I can edit the PR later to address these issues while keeping the current functionality on, and beginning a deprecation move on heatMapLibrary as is already half implemented in the PR.

What do you think?

itsmichaeldiego:
That would be amazing @aviyadeveloper, looking forward to that!

aviyadeveloper:
@itsmichaeldiego Alright i think it's done. :tada:

itsmichaeldiego:
I will release it together with some other fixes asap!

aviyadeveloper:
Nice :blush: If i have some time maybe i pick up more stuff from the issues queue later. Cheers!

itsmichaeldiego:
Merged and released in `2.1.0`, let me know if everything works fine for you!

src/loaders/google_map_loader.js
```diff
@@ -58,13 +58,31 @@ export default (bootstrapURLKeys, heatmapLibrary) => {
       }
     }
 
+    // Support for older version using heatMapLibrary option
+    if (heatMapLibrary) {
```
itsmichaeldiego, line 62: This should be `heatmapLibrary`, got an error.

src/loaders/google_map_loader.js
```diff
@@ -58,13 +58,31 @@ export default (bootstrapURLKeys, heatmapLibrary) => {
...
     const params = Object.keys(bootstrapURLKeys).reduce(
       (r, key) => `${r}&${key}=${bootstrapURLKeys[key]}`,
       ''
     );
-
-    const libraries = heatmapLibrary ? '&libraries=visualization' : '';
-
+    
     $script_(
       `${DEFAULT_URL}${API_PATH}${params}${libraries}`,
```
itsmichaeldiego, line 87: `libraries` here is not defined

itsmichaeldiego (commented):
Just found a bunch of bugs while running your code that didn't notice before, please be more careful in the future

src/loaders/google_map_loader.js
aviyadeveloper: > Just found a bunch of bugs while running your code that didn't notice before, please be more careful in the future

Oh boy yeah, ok will be more diligent later.

itsmichaeldiego:
@aviyadeveloper I have to revert this, as its causing a bug https://github.com/google-map-react/google-map-react/issues/931 and also have many issues, could we work on a proper fix?

Issue #903, opened 2020-08-04
Can't import fitBounds from google-map-react


Venskiy:
I'm trying to update library to the newest release(2.0.4), but I can't import `fitBounds`. When I import `fitBounds` like this: `import { fitBounds } from 'google-map-react';`, webpack will show me this warning: `"export 'fitBounds' was not found in 'google-map-react'`

itsmichaeldiego:
@Venskiy That is odd, I tried myself and it was working, would you please hold on the update till I can check that?

Venskiy:
Thank you for helping.
I will also try to reproduce it on a clean project.

itsmichaeldiego:
@Venskiy Yeah if you could share a live example with https://codesandbox.io/ or similar would be awesome, to see how are you using fitBounds and be able to test it properly

Venskiy:
@itsmichaeldiego 
Here is sandbox example https://codesandbox.io/s/fast-glade-9kouq
I am able to import GoogleMap, but I can't import fitBounds. It's undefined.

itsmichaeldiego:
@Venskiy Awesome, thanks, I will take a look as soon as I can

fracmak:
Would this be part of the problem?

https://github.com/google-map-react/google-map-react/commit/1cae88eb5d8653a1913a0713dc3ce2dbf7214b85#diff-1fdf421c05c1140f6d71444ea2b27638

karldanninger:
@fracmak that would be my guess as well! I'll try to PR a fix later today.

itsmichaeldiego:
@fracmak Yes, indeed that is the problem, exporting the lib files that way resulted in an error with SSR. I rolled it back and checked and it seemed the others libs were being exported too, that was my bad, if @karldanninger doesn't fix it I will asap!

karldanninger:
@itsmichaeldiego here https://github.com/google-map-react/google-map-react/pull/910

Exports all the utilities within the lib directory, including `fitBounds`

adamklepacz:
./utils/helpers/mapHelpers.js
Attempted import error: 'fitBounds' is not exported from 'google-map-react'.

could you guys publish a fix to it?  And finally make it work. Every version from 2.0.0 broke my app so I needed to downgrade to the last stable release which was 1.1.7

itsmichaeldiego:
@adamklepacz You are free to create your own PR, I'll approve it and merge it and release a new version 👍

[...]

Issue #901, opened 2020-07-30
Next.js SSR error after upgrading to 2.0.2


sdornan:
**Describe the bug 🐛**

An error is returned when rendering a map via the server in Next.js. The error is "Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object." 

Version 1.1.7 works fine, so something broke between then and 2.0.2.

**To Reproduce 🔍**

See https://codesandbox.io/s/sleepy-boyd-r2j78?file=/pages/index.js

**Screenshots 🖥**

<img width="989" alt="image" src="https://user-images.githubusercontent.com/563517/88951113-2f8f5180-d25b-11ea-8cad-4ca0deaf82a9.png">

itsmichaeldiego:
Well, it was a big change so I was assuming some problems were going to come our way. Thanks for reporting this. Seems that the issue is only with SSR. I will take a look.

itsmichaeldiego:
This has been fixed and released in `2.0.4`, hope there're no further problems!

@sdornan @jspencersharpe @gbwashleybrown

salonee-vc:
@itsmichaeldiego still facing issue in v2.2.1

PR #923, opened 2020-08-21
Add back lib export

NerdCowboy:
Fixes https://github.com/google-map-react/google-map-react/issues/903

There still may be issues with SSR per https://github.com/google-map-react/google-map-react/issues/903#issuecomment-670234586. I assume that issue is related to the wildcard export, so I manually exported everything instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "prettier": "^2.0.4",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
-    "react-scripts": "^3.4.1"
+    "react-scripts": "^3.4.3"
   },
   "files": [
     "dist"
--- a/src/index.js
+++ b/src/index.js
@@ -1,12 +1,5 @@
 import GoogleMap from './google_map';
 
-export {
-  convertNeSwToNwSe,
-  convertNwSeToNeSw,
-  fitBounds,
-  meters2ScreenPixels,
-  tile2LatLng,
-  latLng2Tile,
-  getTilesIds,
-} from './lib';
+export * from './lib';
+export { GoogleMap };
 export default GoogleMap;
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,4 @@
 import GoogleMap from './google_map';
 
+export * from './lib';
 export default GoogleMap;
--- a/src/index.js
+++ b/src/index.js
@@ -1,13 +1,4 @@
 import GoogleMap from './google_map';
-import {
-  convertNeSwToNwSe,
-  convertNwSeToNeSw,
-  fitBounds,
-  meters2ScreenPixels,
-  tile2LatLng,
-  latLng2Tile,
-  getTilesIds,
-} from './lib';
 
 export {
   convertNeSwToNwSe,
@@ -17,5 +8,5 @@ export {
   tile2LatLng,
   latLng2Tile,
   getTilesIds,
-};
+} from './lib';
 export default GoogleMap;
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,21 @@
 import GoogleMap from './google_map';
+import {
+  convertNeSwToNwSe,
+  convertNwSeToNeSw,
+  fitBounds,
+  meters2ScreenPixels,
+  tile2LatLng,
+  latLng2Tile,
+  getTilesIds,
+} from './lib';
 
-export * from './lib';
+export {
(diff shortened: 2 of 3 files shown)


itsmichaeldiego:
@NerdCowboy Hey! Thanks for the PR, I've tried this but then the issue is https://github.com/google-map-react/google-map-react/issues/901 starts again, the problem is when you export named and default imports with SSR.

You can create a simple project using Next.JS and try it yourself, I haven't found the best way to do it yet

NerdCowboy:
@itsmichaeldiego I did some debugging and found out the issue appears to be using default and named exports in the same file. If I covert the default export `GoogleMap` to a named export, everything works just fine. 

Ignore my commit name on the last push, it does not actually fix the issue entirely, but it does fix it if you consume `GoogleMapReact` as a named export. 

Obviously not ideal as it's a breaking change, but I'll try to look into this issue more later.

itsmichaeldiego:
That would be amazing, thanks @NerdCowboy

itsmichaeldiego:
@NerdCowboy If you look at the latest stable version 1.1.7: https://github.com/google-map-react/google-map-react/tree/1.1.7
both ways were working fine, I think we should check how things were being built back then in order to make a proper fix

NerdCowboy:
The issue appears to be with Microbundle. They don't support and seem to actively discourage using mixed named and default exports per https://github.com/developit/microbundle/issues/306.  I tried the workaround suggested, but it didn't work for me as it looks like the exports var gets renamed when compiled, so it's never exported.

I think React Helmet might have also been bit by this issue, so I'm guessing it's more of a Rollup issue since Microbundle also uses Rollup. https://github.com/nfl/react-helmet/pull/547. They initially only used named exports, but ended using both a named and default export. https://github.com/nfl/react-helmet/pull/395

Not sure what you want to do here. From what I see, here's the options:
- Move off of Microbundle 
- Convert default export to named export (could use both named & default, but wouldn't recommend due to default not working with SSR)
- Force the consumer to pull in the lib methods directly from the sub-directory like it was prior to 2.0

Since 2.0 came with breaking changes and 2.0+ itself has technically been broken since its release for anyone using lib methods, I think the path of least resistance would be to convert the default export to a named one.

For what it's worth, a lot of people are starting to push for only using named exports
https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/
https://basarat.gitbook.io/typescript/main-1/defaultisbad

Let me know what you think

itsmichaeldiego:
Thanks for your research, its super helpful, I didn't know it was a problem with microbundle itself. I did see this problem with rollup though, and I was thinking the only way was to change microbundle to another similar technology.

Let me get this straight, you recommend that we convert the default export to a named one for SSR or in general?

NerdCowboy:
No problem. I'm recommending converting the default export to use a named one only. 

In general, I think the arguments for using only named exports are pretty sound, but I'm not recommending refactoring any other default exports unless you think it's possible you'll want to add named exports from the same file in the future. 

Up to you what you want to do though. Just let me know and I'll update my PR or if you want to tackle it, that's cool too.

karldanninger:
I agree with @NerdCowboy, even though it's not ideal - it may be the way to go moving forward!

itsmichaeldiego:
@karldanninger I agree with you both, I am just worried about the people who are using SSR, they will have to change this if they upgrade, maybe we could add a warning?

itsmichaeldiego:
@NerdCowboy @karldanninger I agree with you both, I am just worried that we're not providing backwards compatibility for those people using SSR today. Instead we're breaking it.

itsmichaeldiego:
I think we should _Force the consumer to pull in the lib methods directly from the sub-directory like it was prior to 2.0_ 

And then, as a second step, move out from microbundle (I can take care of that)

itsmichaeldiego:
@NerdCowboy @karldanninger Hey folks, I've opened a new issue in microbundle: https://github.com/developit/microbundle/issues/712 and I got an answer, will try that and let you know

karldanninger:
You're the man @itsmichaeldiego! 💪💪

I'm glad they gave a useful response on the issue.

itsmichaeldiego:
This was fixed and released in 2.1.3, please let me know if it works!! 🙏

Sorry for the delay! Thanks to both of you!

@karldanninger @NerdCowboy 🙏

PR #950, opened 2020-09-24
Change componentWillReceiveProps to componentDidUpdate

### LINKED ISSUE #944: Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code
**Describe the bug 🐛**
this error is thrown whenever the maps component in loaded, like there is no 'special case' to trigger, it just gets thrown whenever i use the library.

**To Reproduce 🔍**
just use the library and it will be thrown

live example: https://codesandbox.io/s/fervent-bouman-ijrou?file=/src/App.js

**Screenshots 🖥**

image

**Environment:**

 - OS: linux
 - Browser: Brave (chromium based)
 - Version: last version

itsmichaeldiego:
Fixes #944

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -301,122 +301,6 @@ class GoogleMap extends Component {
     }
   }
 
-  // eslint-disable-next-line camelcase
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


itsmichaeldiego (commented):
Seems to be working correctly :shipit:

PR #946, opened 2020-09-22
Use @googlemaps/js-api-loader and fix libraries usage

itsmichaeldiego:
This PR is a cleaned and improved version of the following two PRS:

- #942 feat: use @googlemaps/js-api-loader
- #934 Fix libraries being repeated when using heatmapLibrary

Fixes #922, Resolves #708  

Both PRS had their errors and some unnecessary things, this PR fixes those errors and its a bit more cleaner.

Kindly asking for @aviyadeveloper and @jpoehnelt to review this code as they were the ones who had the initiative, I used some of @jpoehnelt commits and some ideas from @aviyadeveloper's PR #934 

BTW; I will be removing `heatmapLibrary` in the next major version.

PR #942, opened 2020-09-17
feat: use @googlemaps/js-api-loader

jpoehnelt:
Use [@googlemaps/js-api-loader](https://www.npmjs.com/package/@googlemaps/js-api-loader) to dynamically load the Google Maps API.

also fixes #922

PR #955, opened 2020-10-03
Avoid multiple api loads

### LINKED ISSUE #954: google-map-react now load google maps api all the time even if a key  is not provided
**Describe the bug 🐛**

I use this lib, but I manually load the google maps api.
now i just found out that the lib is loaded multiple time:

> You have included the Google Maps JavaScript API multiple times on this page. This may cause unexpected errors.

Using latest version

itsmichaeldiego:
Fix #954

Introduced in #942 and dragged to #946

### DIFF
--- a/src/loaders/google_map_loader.js
+++ b/src/loaders/google_map_loader.js
@@ -1,7 +1,6 @@
 import { Loader } from '@googlemaps/js-api-loader';
 
-let loader_ = null;
-
+let loadPromise_;
 let resolveCustomPromise_;
 
 const _customPromise = new Promise((resolve) => {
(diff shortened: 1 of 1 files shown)


src/loaders/google_map_loader.js
```diff
@@ -50,23 +54,21 @@ export default (bootstrapURLKeys, heatmapLibrary) => {
     throw new Error('google map cannot be loaded outside browser env');
   }
 
-  if (!loader_) {
```
itsmichaeldiego, line 53: @jpoehnelt I feel there's no need to check for this, that doesn't change much, because loader will load another script anyways.

src/loaders/google_map_loader.js
```diff
@@ -16,6 +15,11 @@ export default (bootstrapURLKeys, heatmapLibrary) => {
     return _customPromise;
   }
 
+  // avoid api to be loaded multiple times
+  if (loadPromise_) {
```
itsmichaeldiego, line 19: @jpoehnelt Instead of checking for `!loader_` we return the promise, avoiding the loader to add another script

PR #961, opened 2020-10-12
feature: add support for multiple heatmaps

thunderer199:
Change implementation from one layer to multiple layers. With the compatibility of one layer configuration from previous versions.


itsmichaeldiego:
Hey, this is nice, thanks! Could you provide an example so we can test it out easily?

thunderer199:
You can pass multiple configs to the `heatmap` prop. In other terms in stay unchanged. Not sure where is the best place to add example. 
Maybe I can add some example right in the repo, to show how it works?

itsmichaeldiego:
That would be amazing @thunderer199, thank you!

PR #964, opened 2020-10-19
fix CloneElement Porps

myoneflag:


src/google_map_markers.js
```diff
@@ -263,10 +263,10 @@ export default class GoogleMapMarkers extends Component {
           child.props.lng === undefined
         ) {
           return React.cloneElement(child, {
-            $geoService: this.props.geoService,
-            $onMouseAllow: this._onMouseAllow,
-            $prerender: this.props.prerender,
```
itsmichaeldiego, line 268: Why are we removing the `$` prefix?

itsmichaeldiego (commented):
Wondering what's the issue you're facing and how does this fix it, could you provide more info?

src/google_map_markers.js
myoneflag: Sorry, my misunderstand. I am very interested in this cooperation. Is there any issue let me fix?

myoneflag: Sorry again

PR #972, opened 2020-11-23
add markerPosition prop

IBanda:
added a markerPosition prop.

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -24,6 +24,12 @@ const style = {
   position: 'absolute',
 };
 
+const markerWrapperStyles = {
+  position: 'absolute',
+  top: 0,
+  left: 0,
+};
+
 export default class GoogleMapMarkers extends Component {
   /* eslint-disable react/forbid-prop-types */
   static propTypes = {
@@ -252,6 +258,7 @@ export default class GoogleMapMarkers extends Component {
   render() {
     const mainElementStyle = this.props.style || mainStyle;
     this.dimensionsCache_ = {};
+    this.markerPosition = {};
 
     const markers = React.Children.map(
       this.state.children,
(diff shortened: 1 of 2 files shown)


src/google_map_markers.js
```diff
@@ -320,20 +327,40 @@ export default class GoogleMapMarkers extends Component {
           ...latLng,
         };
 
+        if (child.props.markerPosition) {
+          const markerPosition = child.props.markerPosition;
+          if (typeof markerPosition === 'string') {
+            const [left, top] = markerPosition.split(' ');
+            if (left && top) {
+              this.markerPosition = { transform: `translate(${left},${top})` };
```
itsmichaeldiego, line 335: Let's use `transform-origin` so we can directly pass the property like:

```javascript
transform-origin: ${markerPosition}`
```

src/google_map_markers.js
```diff
@@ -252,6 +258,7 @@ export default class GoogleMapMarkers extends Component {
   render() {
     const mainElementStyle = this.props.style || mainStyle;
     this.dimensionsCache_ = {};
+    this.markerPosition = {};
```
itsmichaeldiego, line 261: I don't think this will be necessary

src/google_map_markers.js
```diff
@@ -24,6 +24,12 @@ const style = {
   position: 'absolute',
 };
 
+const markerWrapperStyles = {
```
itsmichaeldiego, line 27: We already have this, is the variable called `style`, this won't be necessary

src/google_map_markers.js
```diff
@@ -320,20 +327,40 @@ export default class GoogleMapMarkers extends Component {
...
           >
-            {React.cloneElement(child, {
-              $hover: childKey === this.state.hoverKey,
-              $getDimensions: this._getDimensions,
-              $dimensionKey: childKey,
-              $geoService: this.props.geoService,
-              $onMouseAllow: this._onMouseAllow,
-              $prerender: this.props.prerender,
-            })}
+            <div style={{ ...markerWrapperStyles, ...this.markerPosition }}>
```
itsmichaeldiego, line 354: No need of extra div, it already exists above, you should merge the styles to `style` and not create a new one.

itsmichaeldiego (changes requested):
Lets remove `package-lock.json`, we're using yarn here, make sure to use it too!

PR #994, opened 2021-01-30
fix: sometimes changes to the draggable prop don't get propagated

### LINKED ISSUE #978: Map is not beeing updated.
**Describe the bug 🐛**

The map is not beeing updated when I update the props. I have a Map component that receives the map options as props, but when I update the options the map doesn't rerender.

In this example, there is a button at the bottom where it enable and disable the `draggable` and `scrollwheel` options
[Live Example](https://jsbin.com/raqiqaveso/)

tagoro9:
This pull request fixes an issue where a change to the draggable prop would not make the map draggable by not omitting the prop inside the `shouldComponentUpdate` prop.

If we change the prop but there is no other prop change the map behavior won't get updated, as that is done inside the `componentDidUpdate` function.

You can see the issue reproduced in this sandbox: https://codesandbox.io/s/distracted-wright-hrs51?file=/src/App.js

When you make the map draggable only the last map becomes draggable as there is another prop change, while the first 2 can't be dragged (the second one is expected as the code in `componentDidUpdate` omits those).

Fixes #978

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -302,12 +302,9 @@ class GoogleMap extends Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    // draggable does not affect inner components
     return (
-      !shallowEqual(
-        omit(this.props, ['draggable']),
-        omit(nextProps, ['draggable'])
-      ) || !shallowEqual(this.state, nextState)
+      !shallowEqual(this.props, nextProps) ||
+      !shallowEqual(this.state, nextState)
     );
   }
 


src/google_map.js
```diff
@@ -302,12 +302,9 @@ class GoogleMap extends Component {
   }
 
   shouldComponentUpdate(nextProps, nextState) {
-    // draggable does not affect inner components
     return (
-      !shallowEqual(
```
IAmBlackHacker, line 307: seems to be working well for me

PR #1049, opened 2021-07-20
Marker position

IBanda:
PR to introduce the markerPosition prop to position the marker in regards to `transform:translate(x,y)`.
The prop takes a string composed of the X axis value and Y axis value separated by a single white space
The prop can take following values for the X axis;
- left === 0
- center === -50%
- right === -100% 

and the following for the Y axis;

- top === 0
- center === -50%
- bottom === -100%

For example
```
<GoogleMapReact>
<Marker markerPosition="left center"/>
</GoogleMapReact>
```

### DIFF
--- a/src/google_map_markers.js
+++ b/src/google_map_markers.js
@@ -24,6 +24,25 @@ const style = {
   position: 'absolute',
 };
 
+const markerWrapperStyles = {
+  ...style,
+  height: null,
+  width: null,
+};
+
+const markerAxisX = {
+  left: 0,
+  center: '-50%',
+  right: '-100%',
+};
+
+const markerAxisY = {
+  top: 0,
+  center: '-50%',
+  bottom: '-100%',
+};
+const isNotProduction = process.env.NODE_ENV !== 'production';
+
 export default class GoogleMapMarkers extends Component {
   /* eslint-disable react/forbid-prop-types */
   static propTypes = {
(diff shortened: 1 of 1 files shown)


src/google_map_markers.js
```diff
@@ -16,14 +16,25 @@ const mainStyle = {
...
+  center: '-50%',
+  right: '-100%',
+};
+
+const markerAxisY = {
+  top: 0,
+  center: '-50%',
+  bottom: '-100%',
+};
+const isNotProduction = process.env.NODE_ENV !== 'production';
```
itsmichaeldiego, line 36: Why do we need this? :)

IBanda: We are using it to determine whether to console.error

IBanda: Trying to match how proptypes works

itsmichaeldiego: Ok, sounds good! We could call it `isProduction` instead, and then do `if (!isProduction)`

IBanda: Alright :) I will make that update

src/google_map_markers.js
```diff
@@ -320,10 +331,36 @@ export default class GoogleMapMarkers extends Component {
           ...latLng,
         };
 
+        const markerPosition = child.props.markerPosition || 'center center';
```
itsmichaeldiego, line 334: Default should be google map's default, which is bottom center

IBanda: marker positioning

Just for confirmation, is that the same thing referenced in the doc? 👆

itsmichaeldiego: @IBanda We're changing the default position, yes, therefore we have to update the docs

itsmichaeldiego (approved):

PR #1086, opened 2022-01-09
Add support for mapId option.

jakeFeldman:
## Description

This PR adds support for mapId.  By default `mapId` works by adding to the options prop, but we get a warning in the console. This resolves said warning by removing the styles property from the maps option.

I've updated the type defs in the DefinitelyTyped repo as well ([PR here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58072))

```
A Map's styles property cannot be set when a mapId is present.  When a mapId is present Map styles are controlled via the cloud console. Please see documentation at https://developers.google.com/maps/documentation/javascript/styling#cloud_tooling
```

### DIFF
--- a/src/google_map.js
+++ b/src/google_map.js
@@ -388,6 +388,14 @@ class GoogleMap extends Component {
         // remove zoom, center and draggable options as these are managed by google-maps-react
         options = omit(options, ['zoom', 'center', 'draggable']);
 
+        // A Map's styles property cannot be set when a mapId is present. 
+        // When a mapId is present Map styles are controlled via the cloud console. 
+        if ('mapId' in options) {
+          options = omit(options, ['styles']);
+        }
+
+        console.log('options', options);
+
         if ('minZoom' in options) {
           const minZoom = this._computeMinZoom(options.minZoom);
           options.minZoom = _checkMinZoom(options.minZoom, minZoom);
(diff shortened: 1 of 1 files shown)


jakeFeldman:
Hello @itsmichaeldiego, I'm reaching out to see if you could take a look at this PR. It's enabling mapId support inside map options. Thanks in advance!

strarsis:
@jakeFeldman: Awesome! I really need this as I want to use such a map style by map ID!
Could I still use this without this PR, but with a warning in JavaScript console (which would be acceptable until this has been merged and a new release published)?

strarsis:
@jakeFeldman: Also note that the mapId can only be changed (after it was set) by completely resetting the Google Maps instance.

humbroll:
I need this PR as well. I hope it's merged to the main. Is this project not active anymore? It seems like it's been a while since the PR was created.

jakeFeldman:
Not too sure. Doesn't seem like its maintained anymore.

kevin1193:
Waiting for this to be merged! Would be a good feature.

src/google_map.js
```diff
@@ -388,6 +388,14 @@ class GoogleMap extends Component {
         // remove zoom, center and draggable options as these are managed by google-maps-react
         options = omit(options, ['zoom', 'center', 'draggable']);
 
+        // A Map's styles property cannot be set when a mapId is present. 
+        // When a mapId is present Map styles are controlled via the cloud console. 
+        if ('mapId' in options) {
+          options = omit(options, ['styles']);
+        }
+
+        console.log('options', options);
```
agustingabiola, line 397: should remove this

gevvvvv: @jakeFeldman can you resolve this so this can be merged?

jakeFeldman: resolved

dfiedlerx (approved):

PR #1115, opened 2022-05-13
React 18 in peer-dependencies and use new API

tetarchus:


tetarchus (commented):
Merged

lzadhito-kargo:
is this means that right now already support React 18?

PR #1224, opened 2024-05-03
Fix: React 18 markers

BenHodgsonR2R:
Issue: (https://github.com/google-map-react/google-map-react/issues/1223)

This pull request addresses an issue with the markers rendering incorrectly using React `^18`. 

The issues were:
1. StrictMode behaves differently, which exposes an issue with the `dimensionsCache` property in `componentWillUnmount`.
https://react.dev/blog/2022/03/29/react-v18#new-strict-mode-behaviors
2. Batched updates, causing flickering with the markers when the map is repositioned or zoomed. Resolved by opting out of automatic batching for the marker state updates (via `flushSync`).
https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching

PR #1241, opened 2025-01-02
fix: React 18 markers

Pariyawit:
This pull request addresses an issue with the markers rendering incorrectly using React ^18.

https://github.com/google-map-react/google-map-react/pull/1224


itsmichaeldiego:
Deployed in v2.2.5, let me know if it works, thanks

Issue #1238, opened 2024-12-16
Update Package To Supporting React 19


georgs-jegers:
Please update package to supporting React 19

foucdeg:
The first issue the package hits is this breaking change for `ReactDOM.findDOMNode` : 

https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-reactdom-finddomnode

miguelski:
i have the findDOMNode issue too

psimk:
If you **have** to use the package due to company shenanigans & are using `pnpm`. Then you can create a patch using pnpm to remove `findDOMNode` usages, as they are not actually needed for the functionality of the package.

They are leftovers from the days when `google-map-react` used string refs.

RodargDev:
@itsmichaeldiego any chance you can accept the pull request and create a hotfix to solve the issue?

itsmichaeldiego:
Fixed and Deployed in v2.2.4, let me know if it works, thanks

PR #1243, opened 2025-02-24
fix(src/google_map.js): remove unnecessary findDOMNode references

michaeljaltamirano:
Per https://github.com/google-map-react/google-map-react/issues/1238, `findDOMNode` is an artifact of some older code. We can use the `ref` directly.


itsmichaeldiego:
Deployed in v2.2.4, let me know if it works, thanks

